PR #20, opened 2018-09-24
Block php-http/client-common:1.8.0

mxr576:

PR #24, opened 2018-10-31
Add support for HTTPlug 2.0

GeorgosXL:


mxr576:
Wow, HTTPlug 2.0 [has just released](https://travis-ci.org/mxr576/apigee-client-php/builds/448781277). That was quick... :)

Thanks @GeorgosXL for your PR. Please sign the CLA and submit a link to a Travis build that proof your changes does not break any tests _even_ when the online client is being used. (PR tests use mock responses, they do not call the Apigee Edge API.)
These are the variables that you have to configure to run a build in Travis:
https://github.com/apigee/apigee-client-php#unit-tests

Expected link is something like this: https://travis-ci.org/mxr576/apigee-client-php/builds/448781277

mxr576:
Related: https://github.com/php-http/client-common/pull/114

So we should not hurry with this merge of this PR because there are plenty php-http components that version needs to be bumped and for that they need to be compatible with HttpPlug 2.0 (precisely the [finalized PSR-18](https://twitter.com/httplug/status/1057564578786299905)).

Jaesin:
Besides the CLA, I think it would be fair to just do a manual review for a small PR like this one. Of course the dependency that changed should be reviewed as well.

mxr576:
@Jaesin You may think a simple version bump is a "small" change, but we have even seen recently that a simple minor version (reminder, this is a *major version update*) about in PHP API client's dependencies can break the library. See #20. I though it is going to be a simple update but luckily our online test suite that Travis executes revealed that if we would update to the latest major version from the php-client/client-common the PHP API client would stop working.
So we should not make exceptions because setting up the test suite is fairly easy in Travis. If you do not think it is easy enough please provide suggestions about how we should improve our CONTRIBUTING.md. :)

Jaesin:
Ref: https://github.com/php-http/guzzle6-adapter/releases

mxr576:
Client common library's 2.0 version is still in progress: https://github.com/php-http/client-common/issues/115

mxr576:
It seems we can not support 2.x version from php-http packages without releasing a new major version from this library. I am closing this PR and opening an issue where I try to collect all info that I have regarding this.

PR #34, opened 2018-11-27
Added Backward Compatibility promise section

mxr576:


cnovak:
See https://apigeesc.atlassian.net/browse/DRUP-468

mxr576:
@cnovak Could you please check the wording?

README.md
```diff
@@ -158,6 +158,12 @@ folder of your own test data set.
 PS.: Some unit tests cannot be executed when the offline test data set is in use, those are automatically marked as
 skipped.
 
+## Our Backward Compatibility (BC) Promise
+This library follows the [Semantic Versioning](https://semver.org) strategy. It means only major releases (such as 3.0, 4.0 etc.) are allowed to break backward compatibility and we do our best to keep it this way.
```
cnovak, line 162:
```suggestion
This library follows the [Semantic Versioning](https://semver.org) strategy. This means that only major releases (such as 3.0, 4.0 etc.) are allowed to break backward compatibility and we do our best to follow those rules.
```

README.md
```diff
@@ -158,6 +158,12 @@ folder of your own test data set.
...
+The PHP API Client has to be compatible with the latest versions of the supported Apigee Edge APIs.
```
cnovak, line 163: You said "has to be compatible", but I think you meant to declare that we will support the latest version?  See if this is correct:
```suggestion
The PHP API Client will always be compatible with the latest versions of the supported Apigee Edge APIs.
```

README.md
```diff
@@ -158,6 +158,12 @@ folder of your own test data set.
...
+As you can see, almost every class implements at least one interface in this library, therefore, we would like to suggest to rely on interface definitions instead of concrete classes if you build something on the top of this library. Also, prefer encapsulation over inheritance if you extend our classes.
```
cnovak, line 164:
```suggestion
As you can see, almost every class implements at least one interface in this library. Therefore, we would like to suggest that you rely on interface definitions instead of concrete classes if you build something on the top of this library. Also, prefer encapsulation over inheritance if you extend our classes.
```

cnovak (approved):
Changed the wording a small amount as suggestions

PR #43, opened 2019-01-30
DRUP-574 Simulate pagination if CPS is not enabled.

mxr576:
Related: https://github.com/apigee/apigee-edge-drupal/issues/123

This implementation keeps BC and also solves the reported issue. Because PHP API client applies a workaround, a shortcut with non-CPS enabled orgs, we should let developers know what happened. This is the reason why I choose to triegger an E_USER_NOTICE. 
Introducing APIGEE_EDGE_PHP_CLIENT_SUPPRESS_CPS_SIMULATION_NOTICE environment variables seemed to be a good idea because at this moment E_NOTICE errors can not be suppressed if you are using Drupal, it does not matter what do you configure in the PHP ini.
https://www.drupal.org/project/drupal/issues/1267246

Latest test with real Apigee Edge connection: https://travis-ci.org/mxr576/apigee-client-php/builds/486860650


peterserfozo (approved):

mxr576:
Tested it with private cloud build v4.19.01, worked as expected. Both client and Drupal 8 tests passed.
@cnovak Would you like to review this PR and approve it before I merge?

README.md
```diff
@@ -40,6 +40,12 @@ We are planning to add Monetization API support to this library in the near futu
 The [Apigee Monetization APIs](https://apidocs.apigee.com/api-reference/content/monetization-apis) have been added to this library but are
 considered to be an alpha.  If you run into any problems, add an issue to our [GitHub issue queue](https://github.com/apigee/apigee-client-php/issues).
 
+## Edge for Private Cloud
+[Core Persistent Services (CPS)](https://docs.apigee.com/api-platform/reference/cps) is not yet available on Private Cloud installations.
```
cnovak:
```suggestion
[Core Persistent Services (CPS)](https://docs.apigee.com/api-platform/reference/cps) is not available on Private Cloud installations.
```

README.md
```diff
@@ -40,6 +40,12 @@ We are planning to add Monetization API support to this library in the near futu
...
+The PHP API client supports to leverage pagination on listing API endpoints (ex.: [List Developers](https://apidocs.apigee.com/management/apis/get/organizations/%7Borg_name%7D/developers)). If CPS is not available the PHP API client simulates the pagination feature and it triggers an E_USER_NOTICE level error to let developers know that the paginated result is generated by PHP and not the Management API server.
```
cnovak:
```suggestion
The PHP API client supports pagination on listing API endpoints (ex.: [List Developers](https://apidocs.apigee.com/management/apis/get/organizations/%7Borg_name%7D/developers)). If CPS is not available the PHP API client simulates the pagination feature and it triggers an E_USER_NOTICE level error to let developers know that the paginated result is generated by PHP and not the Management API server.
```

cnovak (approved):
General thoughts:

* Paging simulation: I am not sure if paging simulation will work on non CPS orgs. I thought I remember that items are not guaranteed to come back in same order.  I sent an email to persons that can answer this and cc'd you @mxr576 

* I am not sure how the E_USER_NOTICE error surfaces in our Drupal modules. Just want to make sure this message is suppressed without extra user work since non CPS orgs are standard for private cloud users and probably will always be that way. 

*  If a user is using this library against a non CPS org and not trying to use paging there shouldn't be any difference.  If this is true, we could throw exception or error only when trying to use a pager on a nonCPS org (that could also be suppressed).   This way we only expose that something may be behaving differently than expected in that very narrow case.

Since we should solve this ASAP, we can put this PR in now and look at this again if needed.

mxr576:
@cnovak Thanks for the review, I could not commit your suggestion in one commit by using GH's new suggestions feature so I have to commit them in the usual way.

Reacting to your comments:
1. The answer that we got is pretty much unclear and if what it suggest is true we would need to run sorting on the result of a CPS supported listing call and not on the non-CPS ones. Nothing is mentioned about what should be the sorting parameter when objects and not ids are returned. Because we haven't seen any issues yet that would possible caused by the missing sorting I'd leave the code as it is .

2. Because the paginated result on a non-CPS org is generated by PHP and not by the MGMT API server I think we should communicate that to developers. As I wrote in the description I could not find a better way to do that, I had experienced with multiple possible solutions. Setting up environment variables is usually quite easy and straightforward in any environments, so this seemed to be the best solution. Also, it is up to the server admin/devops guy on a project whether E_NOTICE level warnings gets logged in a production environment, if they suppressed this "info" never gets logged in production.
If we get complains about this in the Drupal 8 module we can find a way to provide a configuration UI/automatically suppress this notice.

3. E_USER_NOTICE error is triggered only when someone tries to use the pagination feature on a non-CPS org, if no pager is defined no error is triggered on a  non-CPS org.

Issue #47, opened 2019-03-23
The `keepOriginalStartDate` attribute is not always set on rate plan revisions.


Jaesin:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 2.0.1

#### Actual Behavior
Serializing a rate plan revision without `keepOriginalStartDate` set causes an error.

`Property "keepOriginalStartDate" has not been initialized on instance of Apigee\Edge\Api\Monetization\Entity\StandardRatePlanRevision class. Expected type: "boolean".`

#### Expected Behavior

Serializing a pate plan what was retrieved from the API via the client should not error.

#### Steps to Reproduce the behavior

- First, a rate plan revision that doesn't have the keepOriginalStartDate attribute set to true or false.
- Load the rate plan revision.
- Try to serialize the revision.

Subscribe to a rate plan A, cancel it and then subscribe to a plan A again you should get the error.

#### Proposed Solution

Only return true if keepOriginalStartDate is explicitly set to TRUE.

Jaesin:
This just checks whether the start date is true in `isKeepOriginalStartDate` as opposed to returning the property directly. [keep-original-start-date.patch.txt](https://github.com/apigee/apigee-client-php/files/2998801/keep-original-start-date.patch.txt)

mxr576:
According to the [official API documentation](https://docs.apigee.com/api-platform/monetization/create-future-rate-plans) the `keepOriginalStartDate` property should be always a boolean value in every API responses because if it has not been provided in an API request it falls back to "false".

I did my own quick testing with Postman and as I can see even if the "keepOriginalStartDate" is not being sent to the API, it returns that property with a "false" value as it should.

![image](https://user-images.githubusercontent.com/1755573/54985489-30520c80-4fb1-11e9-9035-f60eb6b83c81.png)

Could you provide some sample raw API calls with API responses to confirm the API may not always return the `keepOriginalStartDate` property in an API response for a rate plan revision?

minnur:
@mxr576 the issue is happening to both developers and the teams and it is happening only to revisions. For instance if you subscribe to a rate plan A and cancel it and then subscribe again to a rate plan A you get the error.

[...]

PR #48, opened 2019-03-25
The `keepOriginalStartDate` attribute is not always set on rate plan

minnur:
Related to https://github.com/apigee/apigee-client-php/issues/47


mxr576:
Closed in favor of #49.

PR #49, opened 2019-03-27
DRUP-662 Keep original start date property can be null

mxr576:
Related issue: #47

This PR closes #48 #52.


mxr576:
@Jaesin I rather prefer what it is in #48, because an "isser" should return a boolean value always. If you would like to return the original value sent (or not sent in this case) by the API we should add a getter to the property that returns the value as-is.

Jaesin (approved):
LGTM. I think it would be a good idea to mark `isKeepOriginalStartDate` as deprecated.

PR #51, opened 2019-04-11
DRUP-487 Added support for Monetization Reports API

mxr576:
Remaining tasks:
- Add test coverage.

Usage example:

```php
<?php
require_once "vendor/autoload.php";
require_once "examples/authentication.inc";

$c = new \Apigee\Edge\Api\Monetization\Controller\ReportDefinitionController($clientFactory->getOrganization(), $clientFactory->getClient());
$criteria = new \Apigee\Edge\Api\Monetization\Structure\Reports\Criteria\BillingReportCriteria('NOVEMBER', 2018);
$criteria->showSummary(TRUE)->showTransactionDetail(TRUE);

try {
    echo $c->generateReport($result->getCriteria()) . PHP_EOL;
} catch (Exception $e) {
    echo $e . PHP_EOL;
}

$def = new \Apigee\Edge\Api\Monetization\Entity\ReportDefinition(['name' => 'PHP API Client Report ' . time(), 'criteria' => $criteria]);
try {
    $c->create($def);
    echo $def->id() . PHP_EOL;
} catch (Exception $e) {
    echo $e . PHP_EOL;
}
```


mxr576:
Git history cleaned, PR is ready for review.
Online tests are also passed: https://travis-ci.org/mxr576/apigee-client-php/builds/519198006

mxr576:
Online test code coverage is 94,48%
https://codecov.io/gh/mxr576/apigee-client-php/branch/DRUP-487

src/Api/Monetization/Controller/CompanyReportDefinitionControllerInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Apigee\Edge\Api\Monetization\Controller;
+
+interface CompanyReportDefinitionControllerInterface extends ReportDefinitionControllerInterface
```
dawehner, line 21: May I ask what's the usecase for an interface without any methods?

tests/Api/Monetization/Denormalizer/ReportCriteriaDenormalizerTest.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * @expectedException \Error
+     */
+    public function testDenormalizeWithAbtractClassNoContext(): void
+    {
+        static::$denormalizer->denormalize((object) [], AbstractCriteria::class, 'json');
+    }
+
+    /**
+     * @expectedException \Apigee\Edge\Exception\InvalidArgumentException
```
dawehner, line 56: There is this article about expected exceptions: https://thephp.cc/news/2016/02/questioning-phpunit-best-practices 

The logic layed out there seems quite sensible, even it matters less here, because there is just a single line in this test.

dawehner (commented):
Finalizing my review. I think in general looks quite good. I've had some comments/questions, not sure it is worth waiting on those.

src/Api/Monetization/Controller/CompanyReportDefinitionControllerInterface.php
mxr576: Interfaces in the PHP API client does not just define the common contract for classes that implements an interface. They also define the expected API endpoint that an implementation should communicate with if it implements the interface. Thanks for this, when you type-hint in a consumer app for an interface, you basically also type-hint to the expected API endpoint as well. (You do not care how a concreate implementation communicates with the API until it communicates with the expected API endpoint and it returns API response in the expected format.)

In this particular case, it depends on the caller (the context) whether or not the implementation of ReportDefinitionControllerInterface is interchangeable with an implementation of a CompanyReportDefinitionControllerInterface or a DeveloperReportDefinitionControllerInterface.

tests/Api/Monetization/Denormalizer/ReportCriteriaDenormalizerTest.php
mxr576: @dawehner Thanks for this article, it was an interesting reading even it is 3 years old.
In my next project, I'll not use the annotation based exception testing, because I do not like that either :) But as you wrote, here it does not really matter which one I use and for consistency reasons, I'd keep the current implementation.

mxr576:
@dawehner Thanks for the review. Please approve the PR if you have no more questions.

mxr576:
@dawehner ping

dawehner (approved):
I wasn't able to see any problems with API compatibility.

dawehner (approved):

Jaesin (approved):
Approved by Dawehner

src/Api/Monetization/Controller/CompanyReportDefinitionControllerInterface.php
dawehner: Oh yeah so you use interfaces sort of a bit like a type alias in a algebratic data type. That makes sense.

PR #52, opened 2019-04-24
[DRUP-662] Allow null to be returned by `isKeepOriginalStartDate`.

### LINKED ISSUE #47: The `keepOriginalStartDate` attribute is not always set on rate plan revisions.

Jaesin:
Fixes https://github.com/apigee/apigee-client-php/issues/47
It was agreed that the SDK doesn't have to make a decision on the value if it is not supplied.

Replaces https://github.com/apigee/apigee-client-php/pull/49/files

### DIFF
--- a/src/Api/Monetization/Entity/Property/KeepOriginalStartDatePropertyAwareTrait.php
+++ b/src/Api/Monetization/Entity/Property/KeepOriginalStartDatePropertyAwareTrait.php
@@ -31,7 +31,7 @@ trait KeepOriginalStartDatePropertyAwareTrait
     /**
      * @inheritdoc
      */
-    public function isKeepOriginalStartDate(): bool
+    public function isKeepOriginalStartDate(): ?bool
     {
         return $this->keepOriginalStartDate;
     }
--- a/src/Api/Monetization/Entity/Property/KeepOriginalStartDatePropertyInterface.php
+++ b/src/Api/Monetization/Entity/Property/KeepOriginalStartDatePropertyInterface.php
@@ -23,7 +23,7 @@ interface KeepOriginalStartDatePropertyInterface
     /**
      * @return bool
      */
-    public function isKeepOriginalStartDate(): bool;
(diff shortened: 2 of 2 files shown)


mxr576:
@Jaesin I rather prefer what it is in #49, because an "isser" should return a boolean value always. If you would like to return the original value sent (or not sent in this case) by the API we should add a getter to the property that returns the value as-is. This is the pattern that I used in other places in this library.
In the serializer component, getters also have higher priority than issers. Therfore if a boolean property has a getter it gets called instead of the isser and if it returns null then the property is not being sent in the API request.

mxr576:
@Jaesin @cnovak If we merge something, let's merge #49. If you agree please close this PR.

Jaesin:
Fair enough, since it was decided that we don't need to make a decision on the value so I guess that means we should add a getter and mark the isser as deprecated.

mxr576:
I do not think that having an isser and a getter is a problem, so why would we mark it as "deprecated"? These are just property accessors, nothing complex. As I mentioned, `isKeepOriginalStartDate` is not the first that has both types of getter.

mxr576:
#49 Got merged

PR #71, opened 2019-10-15
#63 Add Google OAuth authentication

### LINKED ISSUE #63: Add Google OAuth authentication
Use account key to create and sign a JWT (JSON Web Token) that is sent to the Google OAuth 2.0 Authorization Server, which returns an access token. The token can then be used in the authorization header (Authorization: Bearer TOKEN) in making calls to the API. When the token expires, the application repeats the process.

arlina-espinoza:
Closes #63.


arlina-espinoza:
This is ready for review. It adds support for Hybrid authentication, requesting and using a bearer token.
This PR intended to be tested with [apigee/apigee-edge-drupal #268](https://github.com/apigee/apigee-edge-drupal/pull/275). Instructions for setting up a test environment can be found in this gist:
https://gist.github.com/arlina-espinoza/e6b299d64f781e89bf54dc5596be1de2

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Retrieves bearer token and saves it to the token storage.
+     *
+     * @psalm-suppress InvalidCatch - Exception by interface can be caught in PHP >= 7.1.
+     */
+    private function getBearerToken(): void
+    {
+        $now = time();
```
kscheirer, line 106: `time()` shouldn't be used, you can use `\Drupal::time()->getRequestTime();` or an injected service instead, see https://www.drupal.org/node/2785211

mxr576: This is not a Drupal code. I am rather thinking about whether `time()` or `REQUEST_TIME` is the appropriate value here.

kscheirer: doh! you're right, I guess `$_SERVER['REQUEST_TIME']` then unless you really need the exact current time?

mxr576 (changes requested):
Code is failing on code style check, please fix that first and I'll provide a detailed review later, hopefully next week
https://github.com/apigee/apigee-client-php/blob/2.x/CONTRIBUTING.md#code-reviews

arlina-espinoza:
> Code is failing on code style check, please fix that first and I'll provide a detailed review later, hopefully next week
> https://github.com/apigee/apigee-client-php/blob/2.x/CONTRIBUTING.md#code-reviews

I've fixed the code style - but had to disable the copyright/header validation as it would either change the copyright to 2018 on the new files, or change all of the source files to 2019 if I update the header template.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: The docs for the JWT specify that what is needed is `The time the assertion was issued, specified as seconds since 00:00:00 UTC, January 1, 1970.`, so time() should be appropriate here.

mxr576:
>  but had to disable the copyright/header validation as it would either change the copyright to 2018 on the new files, or change all of the source files to 2019 if I update the header template.

Well, should we have the same copyright headers on _all_ files? Which means the date should be the same? This problem bugged me since the beginning of this year... and this year almost ended.
@cnovak

.php_cs.dist
```diff
@@ -30,7 +30,7 @@ return PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'class_definition' => ['singleLine' => false, 'singleItemSingleLine' => true],
         'concat_space' => ['spacing' => 'one'],
-        'header_comment' => ['header' => $header],
+        // 'header_comment' => ['header' => $header],
```
cnovak: Shouldn't we remove this line instead of comment out? Or add comment above explaining why it is commented out

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,175 @@
...
+     *   The service account private key.
+     * @param \Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface $token_storage
+     *   Storage where access token gets saved.
+     * @param string|null $auth_server
+     *   Authentication server.
+     */
+    public function __construct(string $email, string $private_key, OauthTokenStorageInterface $token_storage, ?string $auth_server = null)
+    {
+        $this->email = $email;
+        $this->private_key = $private_key;
```
mxr576: camelCaae

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,175 @@
...
+        $this->tokenStorage = $token_storage;
+        $this->auth_server = $auth_server ?: self::DEFAULT_AUTHORIZATION_SERVER;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function authenticate(RequestInterface $request)
+    {
```
mxr576: DRY! Refactor your code, and if it necessary the Oauth authentication plugin as well to reduce code duplication. 50%+ of this class is a copy-paste of https://github.com/apigee/apigee-client-php/blob/2.x/src/HttpClient/Plugin/Authentication/Oauth.php

mxr576 (changes requested):
tbc tomorrow

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,175 @@
...
+     *   Storage where access token gets saved.
+     * @param string|null $auth_server
+     *   Authentication server.
+     */
+    public function __construct(string $email, string $private_key, OauthTokenStorageInterface $token_storage, ?string $auth_server = null)
+    {
+        $this->email = $email;
+        $this->private_key = $private_key;
+        $this->tokenStorage = $token_storage;
+        $this->auth_server = $auth_server ?: self::DEFAULT_AUTHORIZATION_SERVER;
```
mxr576: camelCase, yes it should be camelCased in the Oauth plugin as well :-(

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Fixed.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Fixed.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: I've added an `AbstractOauth` class, which serves as base for both `HybridOauth2` and `Oauth`. Note: because the `auth_server` property was renamed to `authServer` I've provided a `DeprecatedPropertyTrait` which will serve as fallback - because the apigee_edge Drupal module was accessing the property directly and it needed to have BC.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,129 @@
...
+     * @param string $email
+     *   The service account email.
+     * @param string $privateKey
+     *   The service account private key.
+     * @param \Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface $token_storage
+     *   Storage where access token gets saved.
+     * @param string|null $authServer
+     *   Authentication server.
+     */
+    public function __construct(string $email, string $privateKey, OauthTokenStorageInterface $token_storage, ?string $authServer = null)
```
mxr576: Method parameters should be snake_cased.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,129 @@
...
+        $token = [
+            'iss' => $this->email,
+            'aud' => $this->authServer,
+            'scope' => self::TOKEN_SCOPES,
+            'iat' => $now,
+            // Have the token expire in the maximum allowed time of an hour.
+            'exp' => $now + (60 * 60),
+        ];
+
+        $jwt = JWT::encode($token, $this->privateKey, 'RS256');
```
mxr576: * This algorithm and the openssl_sign() function is only available if the OpenSSL extension enabled, add ext-openssl as a dependency to composer.json.
* encode() can throw a DomainException ("OpenSSL unable to sign data"), inherited from the sign() method.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,129 @@
...
+        $body = [
+            'grant_type' => self::GRANT_TYPE,
+            'assertion' => $jwt,
+        ];
+
+        try {
+            $response = $this->authClient()->post('', http_build_query($body), ['Content-Type' => 'application/x-www-form-urlencoded']);
+            $decoded_response = json_decode((string) $response->getBody(), true);
```
mxr576: Unnecessary variable and it should be camelCased.

src/Utility/DeprecatedPropertyTrait.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+trait DeprecatedPropertyTrait
+{
+    /**
+     * Allows access deprecated/removed properties.
+     */
+    public function __get($name)
+    {
+        if (!isset($this->deprecatedProperties)) {
+            throw new LogicException('The deprecatedProperties property must be defined to use this trait.');
```
mxr576: \LogicException. Do not use classes from the global namespace.

src/Utility/DeprecatedPropertyTrait.php
```diff
@@ -0,0 +1,55 @@
...
+ * To use, include the trait and declare a `deprecatedProperties` property,
+ * with the following format:
+ *
+ * @code
+ * protected $deprecatedProperties = [
+ *     'deprecatedProperty' => 'replacementProperty',
+ * ];
+ * @endcode
+ */
+trait DeprecatedPropertyTrait
```
mxr576, line 34: Hehe, good old friend from Drupal core...
https://github.com/drupal/core/blob/8.7.8/lib/Drupal/Core/DependencyInjection/DeprecatedServicePropertyTrait.php

src/Utility/DeprecatedPropertyTrait.php
```diff
@@ -0,0 +1,55 @@
...
+        if (!isset($this->deprecatedProperties)) {
+            throw new LogicException('The deprecatedProperties property must be defined to use this trait.');
+        }
+
+        if (isset($this->deprecatedProperties[$name])) {
+            $replacement = $this->deprecatedProperties[$name];
+            $class_name = static::class;
+            @trigger_error("The property $name is deprecated in $class_name and will be removed before 3.0.0.", E_USER_DEPRECATED);
+
+            return empty($replacement) ? null : $this->{$replacement};
```
mxr576: Because of `isset($this->deprecatedProperties[$name])`, the `empty($replacement)` check is unnecessary, Or I would rather check if the object actually has a replacement property as a sanity check and if not, I would throw a logic exception again.

tests/Test/HttpClient/Plugin/MockHybridOauth2.php
```diff
@@ -0,0 +1,66 @@
...
+    private $journal;
+    /**
+     * @var \Http\Client\HttpClient
+     */
+    private $httpClient;
+
+    public function __construct(
+        string $email,
+        string $privateKey,
+        \Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface $token_storage,
```
mxr576: Do not use FQCN-s here. (Please fix this in the `MockOauth` test as well.)

tests/Test/HttpClient/Plugin/MockHybridOauth2.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    private $journal;
+    /**
+     * @var \Http\Client\HttpClient
+     */
+    private $httpClient;
+
+    public function __construct(
+        string $email,
+        string $privateKey,
```
mxr576, line 49: Method parameters should be snake_cased.

tests/Test/HttpClient/Plugin/MockOauth.php
```diff
@@ -49,13 +49,13 @@ public function __construct(
         \Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface $token_storage,
         HttpClient $httpClient = null,
         JournalInterface $journal = null,
-        ?string $mfa_token = null,
+        ?string $mfaToken = null,
```
mxr576, line 53: Keep the snake cased parameter names.

tests/HttpClient/Plugin/HybridOauth2Test.php
```diff
@@ -0,0 +1,198 @@
...
+        parent::setUp();
+
+        $this->token_storage = new InMemoryOauthTokenStorage();
+        $this->journal = new TestJournal();
+        $this->client = $this->buildClient();
+    }
+
+    /**
+     * @expectedException \Apigee\Edge\Exception\HybridOauth2AuthenticationException
+     * @expectedException \Apigee\Edge\Exception\OauthAccessTokenAuthenticationException
```
mxr576: Only one exception can be captured like this and the client should throw only one specific exception for a specific edge case.

tests/HttpClient/Plugin/HybridOauth2Test.php
```diff
@@ -0,0 +1,198 @@
...
+    protected static $httpClient;
+
+    /** @var \Apigee\Edge\HttpClient\Utility\JournalInterface */
+    private $journal;
+
+    /** @var \Apigee\Edge\ClientInterface */
+    private $client;
+
+    /** @var \Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface */
+    private $token_storage;
```
mxr576: CamelCase

tests/HttpClient/Plugin/HybridOauth2Test.php
```diff
@@ -0,0 +1,198 @@
...
+            'token_type' => 'Bearer',
+        ];
+        // Auth server returns a new access token.
+        static::$httpClient->addResponse(new Response(200, ['Content-Type' => 'application/json'], json_encode((object) $body)));
+
+        // Successful response to an authorised API call from API server.
+        static::$httpClient->addResponse(new Response(200));
+
+        // Another successful response to an authorised API call from API server.
+        static::$httpClient->addResponse(new Response(200));
```
mxr576, line 111: Why do you have to test the same thing twice?

tests/HttpClient/Plugin/HybridOauth2Test.php
```diff
@@ -0,0 +1,198 @@
...
+
+        $body = [
+            'access_token' => 'access_token',
+            'expires_in' => 60,
+            'token_type' => 'Bearer',
+        ];
+        // Auth server returns a new access token.
+        static::$httpClient->addResponse(new Response(200, ['Content-Type' => 'application/json'], json_encode((object) $body)));
```
mxr576, line 122: Again, you are testing the same thing. (Or maybe you should move these setup stuff closer to the actual assertions.)

tests/HttpClient/Plugin/HybridOauth2Test.php
```diff
@@ -0,0 +1,198 @@
...
+
+        // Successful response to an authorised API call from API server.
+        static::$httpClient->addResponse(new Response(200));
+
+        $this->client->get('');
+        $this->client->get('');
+
+        // Mark token as expired so the following API call requests a new token.
+        $this->token_storage->markExpired();
```
mxr576: Instead of marking the token expired directly in the storage, mock an API response that would indicate that the token expired, therefore the client has to ask a new one. What is a response for a request on the hybrid server which was authenticated with an expired token? HTTP 401?
https://github.com/apigee/apigee-client-php/blob/2.x/tests/HttpClient/Plugin/OauthTest.php#L94-L112

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->privateKey = $privateKey;
+        parent::__construct($token_storage, $authServer ?: self::DEFAULT_AUTHORIZATION_SERVER);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function authClient(): ClientInterface
+    {
+        return new Client(new NullAuthentication(), $this->authServer);
```
mxr576, line 96: Well, there are two possible solutions here:
1. What you did, introducing a new NullAthentication plugin.
2. Make the first parameter optional. As I can see, the client's implementation can already handle that. `\Apigee\Edge\Client::getDefaultPlugins()`.

If we stick the current solution, `\Apigee\Edge\Tests\Test\HttpClient\Plugin\NullAuthentication` can be removed (there is no BC for test files) because it is a duplicate. If we would go with the 2nd option then we would remove an important constraint for non-hibrid orgs. On those orgs you must provide an authentication plugin (usually basic auth) to be able to call an endpoint, let it be an API endpoint or the OAuth authorization server.)

mxr576 (changes requested):

src/Utility/DeprecatedPropertyTrait.php
arlina-espinoza: Fixed.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Symfony coding standard says parameters should be camelCase for [arguments](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions).

src/Utility/DeprecatedPropertyTrait.php
arlina-espinoza: Agreed, updated to throw a logic exception if the replacement property wasn't declared.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Fixed.

tests/HttpClient/Plugin/HybridOauth2Test.php
arlina-espinoza: Fixed.

tests/Test/HttpClient/Plugin/MockHybridOauth2.php
arlina-espinoza: Fixed in both places.

tests/Test/HttpClient/Plugin/MockHybridOauth2.php
arlina-espinoza: Symfony coding standard says parameters should be camelCase for [arguments](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions).

tests/Test/HttpClient/Plugin/MockOauth.php
arlina-espinoza: Symfony coding standard says parameters should be camelCase for [arguments](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions).

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Let's keep (1), as it keeps the constraint for the non-hybrid orgs. I've removed the `NullAuthentication` class from the tests and replaced it with this one.

tests/HttpClient/Plugin/HybridOauth2Test.php
arlina-espinoza: Fixed.

tests/HttpClient/Plugin/HybridOauth2Test.php
arlina-espinoza: I've added more comments that describe what we are testing and how each of those responses are used, hopefully it's clearer: https://github.com/arlina-espinoza/apigee-client-php/blob/0f8e71ed50bb2082c68ed9534e2a44af64ac211a/tests/HttpClient/Plugin/HybridOauth2Test.php#L96-L168

tests/HttpClient/Plugin/HybridOauth2Test.php
arlina-espinoza: See https://github.com/apigee/apigee-client-php/pull/71#discussion_r341357331

tests/HttpClient/Plugin/HybridOauth2Test.php
arlina-espinoza: Good suggestion, I've updated it to use the 401 response, which is what the API returns for an expired token.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Fixed.

src/HttpClient/Plugin/Authentication/AbstractOauth.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * @var string
+     */
+    protected $authServer;
+
+    /**
+     * @inheritdoc
+     */
+    protected $deprecatedProperties = ['auth_server' => 'authServer'];
```
cnovak, line 47: The authServer attribute is deprecated, but I do not see any deprecation phpdoc on this class, or any information on what should be used instead. Also, the new hybrid code uses this attribute, is that code being changed?

cnovak (changes requested):
The tests fail due to `Error: Class 'Apigee\Edge\Tests\Test\HttpClient\Plugin\NullAuthentication' not found` on multiple classes.

src/HttpClient/Plugin/Authentication/AbstractOauth.php
arlina-espinoza: The corresponding [PR in the Apigee Edge module](https://github.com/apigee/apigee-edge-drupal/pull/275) has been updated to match. Also, a deprecation notice is thrown in the `DeprecatedPropertyTrait`. As we are only renaming protected variables (and providing backwards compatibility) and not changing an interface, it should not be a big deal.

.php_cs.dist
arlina-espinoza: Yes, I had it commented out while we dealt with #79 , but now I have removed it.

arlina-espinoza:
> The tests fail due to `Error: Class 'Apigee\Edge\Tests\Test\HttpClient\Plugin\NullAuthentication' not found` on multiple classes.

This has been fixed @cnovak :)

.php_cs.dist
```diff
@@ -30,7 +30,6 @@ return PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'class_definition' => ['singleLine' => false, 'singleItemSingleLine' => true],
         'concat_space' => ['spacing' => 'one'],
-        'header_comment' => ['header' => $header],
```
mxr576: Disable this is a separate PR, merge it, because it is currently blocking several PRs. (Leave a comment with a reference to #81 in code)

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,141 @@
...
+
+        try {
+            $response = $this->authClient()->post('', http_build_query($body), ['Content-Type' => 'application/x-www-form-urlencoded']);
+            $decodedResponse = json_decode((string) $response->getBody(), true);
+            $this->tokenStorage->saveToken($decodedResponse);
+        } catch (OauthRefreshTokenExpiredException $e) {
+            // Clear data in token storage because refresh token has expired.
+            $this->tokenStorage->removeToken();
+            // Try to automatically get a new access token.
+            $this->getAccessToken();
```
mxr576: This could lead to infinite recursion, isn't it? The retry plugin handles when a request should be retried or not, this is not the task of the auth plugin.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
```diff
@@ -0,0 +1,141 @@
...
+            'aud' => $this->authServer,
+            'scope' => self::TOKEN_SCOPES,
+            'iat' => $now,
+            // Have the token expire in the maximum allowed time of an hour.
+            'exp' => $now + (60 * 60),
+        ];
+
+        try {
+            $jwt = JWT::encode($token, $this->privateKey, 'RS256');
+        } catch (DomainException $e) {
```
mxr576, line 118: `\DomainException` JWT lib does not have its own namespace reserved that we could use to identify exceptions thrown by the lib :(

src/HttpClient/Plugin/Authentication/NullAuthentication.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
```
mxr576: So even Github thinks this is just a move of a file, should we change the Copyright date on the file in this case as well? Based on what @cnovak wrote earlier in a different issue we should not.

src/Utility/DeprecatedPropertyTrait.php
```diff
@@ -0,0 +1,57 @@
...
+        if (!isset($this->deprecatedProperties)) {
+            throw new \LogicException('The deprecatedProperties property must be defined to use this trait.');
+        }
+
+        if (isset($this->deprecatedProperties[$name])) {
+            $replacement = $this->deprecatedProperties[$name];
+            $className = static::class;
+            @trigger_error("The property $name is deprecated in $className and will be removed before 3.0.0.", E_USER_DEPRECATED);
+
+            if (empty($replacement)) {
```
mxr576: 1. I suggested checking whether the replacement property actually exists in the current class (or its parents), if not, the developer probably made a typo or something. Please validate that and throw an exception before triggering an error.
2. I do not have a concrete example in my head at this moment, but probably it could happen that a property gets removed in the next version completely, no replacement. This is what your current implementation handles and in that case you should also trigger an error instead of throwing an exception to let the developer know it relies on a deprecated property with no replacement.

mxr576 (changes requested):
Checked 80% of the code, good improvements, there are still some todos.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
mxr576: Ahh right, I mixed the code style with Drupal :(

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: Yeah, this is what is thrown by JWT if unable to encode the token :woman_shrugging:

src/HttpClient/Plugin/Authentication/NullAuthentication.php
arlina-espinoza: Yeah, it had 2019 because originally I had both classes, but now that we've removed the one in the tests, it makes sense that it is the same file that was moved.

.php_cs.dist
arlina-espinoza: Done! (https://github.com/apigee/apigee-client-php/pull/83)

src/Utility/DeprecatedPropertyTrait.php
arlina-espinoza: @mxr576 I've restructured the `DeprecatedPropertyTrait` class a bit. It checks if the replacement property is defined and actually exists, otherwise it logs the deprecation error (after all, a deprecated property is being accessed), and then throws an exception (because php would be trying to access an undefined property).
Also added a [test](https://github.com/arlina-espinoza/apigee-client-php/blob/63-add-hybrid-authentication/tests/PropertyAccess/DeprecatedPropertyTraitTest.php) for this class.

src/HttpClient/Plugin/Authentication/HybridOauth2.php
arlina-espinoza: I've been digging deeper into this and found that we can remove this part, as this implementation does not use refresh tokens, so this exception would never be reached.

cnovak (approved):
LGTM, but I see that integration tests are failing.

PR #74, opened 2019-10-25
[#65] Deprecate "createdBy" and "lastModifiedBy" properties

### LINKED ISSUE #65: Deprecate  "createdBy" and "lastModifiedBy" properties
Deprecate  "createdBy" and "lastModifiedBy" properties, as they are not supported (returned) by Hybrid - applications using this client library should not rely on specific platform implementations.

arlina-espinoza:
Closes #65 . Deprecate "createdBy" and "lastModifiedBy" properties, as they are not being used by the Drupal Edge module, and are not supported by Hybrid.
This PR adds deprecation comments and removes the properties from the test mock data as well.


src/Entity/CommonEntityPropertiesAwareTrait.php
```diff
@@ -38,6 +38,10 @@ trait CommonEntityPropertiesAwareTrait
     /**
      * Email address of the organization user who created the entity.
      *
+     * @deprecated Not used in Hybrid orgs.
```
mxr576: Can we used a similar, standardized format like Drupal 8 introduced which tells a deveoper which version XY was deprecated and when it will be removed? Also, if there is any replacement for the class/method/etc a developer can get information about it?

https://github.com/drupal/core/blob/b3273887e822d745ed277ce36cd07dc56e795d83/includes/file.inc#L49

mxr576 (changes requested):
Just because these properties are deprecated I would not remove them from the JSON files and the tests because non-hybrid orgs still support them and we should have test coverage for these.

arlina-espinoza:
> Just because these properties are deprecated I would not remove them from the JSON files and the tests because non-hybrid orgs still support them and we should have test coverage for these.

You are right, I've added them back.

src/Entity/CommonEntityPropertiesAwareTrait.php
arlina-espinoza: I've changed it to:

```
@deprecated in 2.0.4, will be removed before 3.0.0. Unsupported in Hybrid.
@see https://github.com/apigee/apigee-client-php/issues/65
```

_2.0.4 should be the next release._

minnur (approved):
This looks good to me. TravisCI is failing, looks like it is not related to this PR.

src/Entity/CommonEntityPropertiesAwareTrait.php
mxr576: > Unsupported in Hybrid.

What does it mean exactly? What happens if I call these methods on an object returned from a hybrid org? Do they return an empty string?

arlina-espinoza: @mxr576 The fields are empty so calling the methods would return NULL, as they are never initialized.

PR #75, opened 2019-10-25
[#69] Add isHybrid() helper method to organization

### LINKED ISSUE #69: Add validation to check for hybrid org
Add an Organization::isHybrid() helper method that checks the organization property “features.hybrid.enabled”.

arlina-espinoza:
Closes #69.

### DIFF
--- a/src/Api/Management/Entity/OrganizationInterface.php
+++ b/src/Api/Management/Entity/OrganizationInterface.php
@@ -54,4 +54,11 @@ interface OrganizationInterface extends
      * @return array
      */
     public function getTypes(): array;
+
+    /**
+     * Returns TRUE if the organization is a Hybrid org.
+     *
+     * @return bool
+     */
+    public function isHybrid(): bool;
 }
--- a/src/Api/Management/Entity/Organization.php
+++ b/src/Api/Management/Entity/Organization.php
@@ -84,4 +84,14 @@ class Organization extends Entity implements OrganizationInterface
     {
         return self::TYPES;
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function isHybrid(): bool
+    {
+        $isHybrid = $this->getPropertyValue('features.hybrid.enabled');
+
+        return (bool) $isHybrid;
+    }
 }
--- a/tests/Api/Management/Controller/OrganizationControllerTest.php
+++ b/tests/Api/Management/Controller/OrganizationControllerTest.php
@@ -49,10 +49,30 @@ class OrganizationControllerTest extends ControllerTestBase
         $this->assertEquals(['prod', 'test'], $entity->getEnvironments());
         $this->assertTrue($entity->hasProperty('self.service.virtual.host.enabled'));
         $this->assertEquals('true', $entity->getPropertyValue('features.isCpsEnabled'));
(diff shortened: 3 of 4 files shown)


minnur (approved):
Looks good to me.  TravisCI if failing @arlina-espinoza  I am not 100% sure why it is failing but it seems like it is not related to this PR.

src/Api/Management/Entity/OrganizationInterface.php
```diff
@@ -54,4 +54,11 @@ public function setType(string $type): void;
      * @return array
      */
     public function getTypes(): array;
+
+    /**
+     * Returns TRUE if the organization is a Hybrid org.
+     *
+     * @return bool
+     */
+    public function isHybrid(): bool;
```
mxr576, line 63: Well, technically this is a BC breaking change because whoever implemented its own implementation for this interface it has to provide an implementation for this new method as well. Until s/he does not do that, its code will fail with fatal errors. IOW, users cannot just update to the new minor release because it may break their existing code. This is against semantic versioning.

This could be solved in a non-breaking way as well: Introduce a new class (`class HybridOrganisation implements OrganizationInterface`), the object type would clearly indicate if an org is a hyrid org. The organization JSON serializer would return an instance of this new class instead of an `Organisation` if in the API response contains
```json
            {
                "name": "features.hybrid.enabled",
                "value": "true"
            },
```

(This is how developer- and company serializations are also solved, the type of the object clearly indicates the type of the legal entity.)

mxr576 (changes requested):
BC break

src/Api/Management/Entity/OrganizationInterface.php
mxr576: Btw, as their is no "shortcut" in the PHP API client to identify whether the org is monetizaton enabled, because it could be figured out with a trivial one liner code line this, probably this method should not be added either.  Any enabled feature on an Apigee (org) instance can be identified from the return value of `$org->getProperty()`. If you add a new feature flag, like you did with hybrid org, the implementation in the PHP API client does not need the change.

arlina-espinoza: I like the first approach suggested, as it seems to me this "property" is an internal implementation detail that users of this library should not need to know, and it keeps with the way the other entities are implemented. I'll close this PR in favor of this [other one #80](https://github.com/apigee/apigee-client-php/pull/80) that implements it that way.

arlina-espinoza: I like the first approach suggested, as it seems to me this "property" is an internal implementation detail that users of this library should not need to know, and it keeps with the way the other entities are implemented. I'll close this PR in favor of this [other one #80](https://github.com/apigee/apigee-client-php/pull/80) that implements it that way.

arlina-espinoza:
I like the first approach suggested in https://github.com/apigee/apigee-client-php/pull/75#discussion_r340638941, as it seems to me this "property" is an internal implementation detail that users of this library should not need to know, and it keeps with the way the other entities are implemented. I'll close this PR in favor of this [other one #80](https://github.com/apigee/apigee-client-php/pull/80) that implements it that way.

PR #80, opened 2019-10-31
[#69] Hybrid orgs have their own interface and class.

### LINKED ISSUE #69: Add validation to check for hybrid org

arlina-espinoza:
Closes #69, using the approach suggested in https://github.com/apigee/apigee-client-php/pull/75#discussion_r340638941.

### DIFF
--- a/src/Api/Management/Controller/OrganizationController.php
+++ b/src/Api/Management/Controller/OrganizationController.php
@@ -18,7 +18,7 @@
 
 namespace Apigee\Edge\Api\Management\Controller;
 
-use Apigee\Edge\Api\Management\Entity\Organization;
+use Apigee\Edge\Api\Management\Entity\OrganizationInterface;
 use Apigee\Edge\Api\Management\Serializer\OrganizationSerializer;
 use Apigee\Edge\ClientInterface;
 use Apigee\Edge\Controller\AbstractEntityController;
@@ -64,6 +64,8 @@ class OrganizationController extends AbstractEntityController implements Organiz
      */
     protected function getEntityClass(): string
     {
-        return Organization::class;
+        // It could be either a organization - or a hybrid organization. The OrganizationDenormalizer can automatically
+        // decide which one should be used.
+        return OrganizationInterface::class;
     }
 }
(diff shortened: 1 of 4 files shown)


mxr576:
> I like the first approach suggested in #75 (comment), as it seems to me this "property" is an internal implementation detail that users of this library should not need to know, and it keeps with the way the other entities are implemented. I'll close this PR in favor of this other one #80 that implements it that way.

Source: https://github.com/apigee/apigee-client-php/pull/75#discussion_r341379847,

Well, I only realized after I submitted my second comment to that thread, what is the problem with the first approach that I described earlier. Let's continue from here: https://github.com/apigee/apigee-client-php/pull/75#discussion_r340734838

> as their is no "shortcut" in the PHP API client to identify whether the org is monetizaton enabled

Just as there is no getter or organization subclass for CPS enabled orgs. So if we would go with this "subclasses for all Edge features" approach, should we introduce a new subclass for MonetizationOrganisaton, CpsEnabledOrganisation, what if an org is both CPS- and Monetization enabled? An there are other "features" that we do not handle at this moment...

The possible property names and values are the implementation details of Apigee Edge, they will probably change in the future.

> Any enabled feature on an Apigee (org) instance can be identified from the return value of $org->getProperty(). If you add a new feature flag, like you did with hybrid org, the implementation in the PHP API client does not need the change.

I actually like the current approach in the m10 module where you only retrieve and work with the related Apgiee Edge property where it is actually needed. https://github.com/apigee/apigee-m10n-drupal/blob/bfb863076686c54f1ff5a5d5877fb4590b490990/src/Monetization.php#L126-L154 Can't we just do the same in the Apigee Edge method and introduce a helper method/function for this check only where we actually need this information?

mxr576:
3rd possible option, if you really would like to provide shortcuts for figuring out what features are enabled on an Org in the PHP API client level, could be introducing a new `OrganisationFeatureChecker` class with static methods that could return boolean values of _some_ handled features at this moment, like CPS, monetization, hybrid. The number of methods and supported features could be extended in time. (Compared with the original approach in #75 that would lead to BC break.)

```php

namespace Apigee\Edge\Utility;

final class OrganisationFeatureChecker {

public static function isCpsEnabled(OrganisationInterface $org): bool {
// TODO
}

public static function isHybridEnabled(OrganisationInterface $org): bool {
// TODO
}

// ...
}

```

arlina-espinoza:
Closed via #82.

Issue #81, opened 2019-11-04
Add a copyright validation to the code style tools


arlina-espinoza:
Because we need to keep the copyright year of a file pinned, and don't change it if the file is edited/moved/copied, we had to remove the `header_comment` fixer from the PHP-CS-Fixer rules (see https://github.com/apigee/apigee-client-php/issues/79#issuecomment-548592357).

It would be good to have a code fixer/sniff that adds this back in a way that we it adds a copyright header with the current year to new files, and ensures the header exists on other files, or at least verifies that all files have a copyright header.

Related: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4625

Issue #79, opened 2019-10-31
Actualize copyright headers


mxr576:
Moved conversation from an active PR: https://github.com/apigee/apigee-client-php/pull/71#issuecomment-547275120

@cnovak please share your thoughts about it. I would not drop the license header plugin unless it is actually a legal requirement having different copyright headers on older files.

cnovak:
The files need:

> 1. the word Copyright
> 2. the year the file was created in full numeric form (e.g., 2010); don't change this if you edit, move, or copy the file.
> 3. the author (Google LLC)
> 
> All of these must be together on a single line.

mxr576:
> the year the file was created
> don't change this if you edit, move, or copy the file.

Ahh, so this is a tricky part... I would check how complicated would it be convincing PHPCSFixer about not changing dates in header comments. It seems there is no OOTB option for that yet, although it should be possible to write a custom PHPCSFixer fixer based on the header comment fixer that adds a copyright header with the current year to new files, and ensures the header exists on other files.

https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4625

arlina-espinoza:
Closing this issue in favor of this other one that follows up on adding the validation back in a new way: https://github.com/apigee/apigee-client-php/issues/81

PR #83, opened 2019-11-05
[#81] Remove header_comment validation.

arlina-espinoza:
This PR removes the `header_comment` validation because of the reasons explained in https://github.com/apigee/apigee-client-php/issues/79#issuecomment-548592357.

Related to #81 .


cnovak (approved):
LGTM

PR #84, opened 2019-11-07
[#66] Changing criteria to check if pagination is enabled.

### LINKED ISSUE #66: Remove check on “features.isCpsEnabled” to decide on pagination.
Pagination should be default for lists in hybrid.

arlina-espinoza:
Resolves #66 . Adds helper `OrganizationFeatures::isPaginationEnabled()` method to decide if pagination should be used for an org.

### DIFF
--- a/src/Utility/OrganizationFeatures.php
+++ b/src/Utility/OrganizationFeatures.php
@@ -35,6 +35,18 @@ final class OrganizationFeatures
     }
 
     /**
+     * Checks whether pagination is enabled on an organization.
+     *
+     * @param \Apigee\Edge\Api\Management\Entity\OrganizationInterface $org
+     *
+     * @return bool
+     */
+    public static function isPaginationEnabled(OrganizationInterface $org): bool
+    {
+        return static::isCpsEnabled($org) || static::isHybridEnabled($org);
+    }
+
+    /**
      * Checks whether hybrid feature is enabled on an organization.
      *
      * @param \Apigee\Edge\Api\Management\Entity\OrganizationInterface $org
--- a/src/Utility/OrganizationFeatures.php
+++ b/src/Utility/OrganizationFeatures.php
@@ -41,7 +41,7 @@ final class OrganizationFeatures
      *
      * @return bool
      */
-    public static function isPaginationEnabled(OrganizationInterface $org): bool
+    public static function isPaginationAvailable(OrganizationInterface $org): bool
     {
         return static::isCpsEnabled($org) || static::isHybridEnabled($org);
     }
--- a/tests/Utility/OrganizationFeaturesTest.php
+++ b/tests/Utility/OrganizationFeaturesTest.php
@@ -30,6 +30,8 @@ use PHPUnit\Framework\TestCase;
 class OrganizationFeaturesTest extends TestCase
 {
     /**
+     * Tests simple properties on the organization.
+     *
      * @dataProvider featurePropertyValueProvider
      */
     public function testOrganizationFeatures(?string $propertyValue, bool $expectedResult): void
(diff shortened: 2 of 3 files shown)


tests/Utility/OrganizationFeaturesTest.php
```diff
@@ -42,6 +44,34 @@ public function testOrganizationFeatures(?string $propertyValue, bool $expectedR
...
+        $organization->method('getPropertyValue')->will($this->returnValueMap([
+            ['features.isCpsEnabled', null],
+            ['features.hybrid.enabled', 'true'],
+        ]));
+        $this->assertTrue(OrganizationFeatures::isPaginationEnabled($organization));
+
+        $organization = $this->getMockBuilder(OrganizationInterface::class)->getMock();
+        $organization->method('getPropertyValue')->will($this->returnValueMap([
+            ['features.isCpsEnabled', 'true'],
+            ['features.hybrid.enabled', null],
```
cnovak: It would be a more realistic test if you remove this line since an org that is not hybrid will not return the `features.hybrid.enabled` at all:

```
GET https://api.enterprise.apigee.com/v1/o/myorg

{
  ...
  "name": "myorg",
  "properties": {
    "property": [
      {
        "name": "features.isSmbOrganization",
        "value": "true"
      },
      {
        "name": "features.isCpsEnabled",
        "value": "true"
      }
    ]
  },
  "type": "paid"
}
```

cnovak (approved):
Will approve, but consider my comment about test, and make sure to fix tests before we merge into dev.

src/Utility/OrganizationFeatures.php
```diff
@@ -34,6 +34,18 @@ public static function isCpsEnabled(OrganizationInterface $org): bool
...
     }
 
+    /**
+     * Checks whether pagination is enabled on an organization.
+     *
+     * @param \Apigee\Edge\Api\Management\Entity\OrganizationInterface $org
+     *
+     * @return bool
+     */
+    public static function isPaginationEnabled(OrganizationInterface $org): bool
```
mxr576: Isn't it rather `isPaginationAvailable()`? It is not an enabled feature of an organization in this sense, because the information is not directly available on the organization object, the result is calculated from multiple enabled features as I can see.

tests/Utility/OrganizationFeaturesTest.php
```diff
@@ -42,6 +44,34 @@ public function testOrganizationFeatures(?string $propertyValue, bool $expectedR
...
+        ]));
+        $this->assertTrue(OrganizationFeatures::isPaginationEnabled($organization));
+
+        $organization = $this->getMockBuilder(OrganizationInterface::class)->getMock();
+        $organization->method('getPropertyValue')->will($this->returnValueMap([
+            ['features.isCpsEnabled', 'false'],
```
mxr576: Please introduce a data provider, just like I did for `testOrganizationFeatures`, that will reduce the code duplication in the test method.

mxr576 (changes requested):

tests/Utility/OrganizationFeaturesTest.php
arlina-espinoza: Refactored test :+1:

src/Utility/OrganizationFeatures.php
arlina-espinoza: Agreed, I've renamed the method.

tests/Utility/OrganizationFeaturesTest.php
arlina-espinoza: `OrganizationFeatures::isFeatureEnabled()` will read the `features.hybrid.enabled` property on the organization; if it's not set, it returns null, which is what the dataset is providing.

arlina-espinoza:
@cnovak / @mxr576 This is ready for another review :)

src/Utility/OrganizationFeatures.php
cnovak: Good call, makes things easier to read w/proper names.

tests/Utility/OrganizationFeaturesTest.php
cnovak: Still good to have our test data mimicking the real data as close as possible. What if `isFeatureEnabled()` changes?

minnur (approved):
Looks good to me.

cnovak (commented):
Approved, but think about my one comment to see if it makes sense

tests/Utility/OrganizationFeaturesTest.php
arlina-espinoza: @cnovak On a non-hybrid org, the property `features.hybrid.enabled` is not present, so `$organization->getPropertyValue('features.hybrid.enabled')` would return null, which is what the data provider is returning:
```
// Format: ['features.isCpsEnabled', 'features.hybrid.enabled', $paginationAvailable, $message]
public function paginationAvailableValueProvider(): array
    {
        return [
            [null, 'true', true, 'Hybrid orgs should have pagination.'],
            ['true', null, true, 'CPS enabled orgs should have pagination.'],
            ['false', null, false, 'Non-Hybrid org without CPS should not have pagination.'],
        ];
    }
```

PR #85, opened 2019-11-07
[#67] Hybrid compatibility with "expand=false".

### LINKED ISSUE #67: Confirm if "Listing" operations work as previous edge version.
Test out the behavior of the query parameter “expand=false” - it might return the same format as current Edge.

arlina-espinoza:
Resolves #67. Hybrid returns a different response when using "expand=false": 
```
{
  "proxies": [
    {
      "name": "helloworld"
    },
    {
      "name": "weather"
    }
  ]
}
```
While current Edge returns: 
```
[
  "helloworld",
  "weather"
]
```

This PR formats Hybrid responses in a compatible way.

### DIFF
--- a/src/Controller/NonPaginatedEntityIdListingControllerTrait.php
+++ b/src/Controller/NonPaginatedEntityIdListingControllerTrait.php
@@ -18,6 +18,7 @@
 
 namespace Apigee\Edge\Controller;
 
+use Apigee\Edge\ClientInterface;
 use Psr\Http\Message\ResponseInterface;
 
 /**
@@ -40,8 +41,9 @@ trait NonPaginatedEntityIdListingControllerTrait
         ];
         $uri = $this->getBaseEndpointUri()->withQuery(http_build_query($query_params));
         $response = $this->getClient()->get($uri);
+        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT == $this->getClient()->getEndpoint());
 
-        return $this->responseToArray($response);
+        return $this->responseToArray($response, $expandCompatibility);
     }
 
     /**
--- a/src/Controller/NonPaginatedEntityIdListingControllerTrait.php
+++ b/src/Controller/NonPaginatedEntityIdListingControllerTrait.php
@@ -41,7 +41,7 @@ trait NonPaginatedEntityIdListingControllerTrait
         ];
         $uri = $this->getBaseEndpointUri()->withQuery(http_build_query($query_params));
         $response = $this->getClient()->get($uri);
-        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT == $this->getClient()->getEndpoint());
+        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT === $this->getClient()->getEndpoint());
 
         return $this->responseToArray($response, $expandCompatibility);
     }
--- a/src/Controller/NonPaginatedEntityIdListingControllerTrait.php
+++ b/src/Controller/NonPaginatedEntityIdListingControllerTrait.php
@@ -49,5 +49,5 @@ trait NonPaginatedEntityIdListingControllerTrait
     /**
      * @inheritdoc
      */
-    abstract protected function responseToArray(ResponseInterface $response): array;
+    abstract protected function responseToArray(ResponseInterface $response, bool $expandCompatibility = false): array;
 }
--- a/src/Controller/NonPaginatedEntityListingControllerTrait.php
+++ b/src/Controller/NonPaginatedEntityListingControllerTrait.php
@@ -50,7 +50,7 @@ trait NonPaginatedEntityListingControllerTrait
     /**
      * @inheritdoc
      */
-    abstract protected function responseToArray(ResponseInterface $response): array;
+    abstract protected function responseToArray(ResponseInterface $response, bool $expandCompatibility = false): array;
 
     /**
      * @inheritdoc
--- a/src/Controller/PaginationHelperTrait.php
+++ b/src/Controller/PaginationHelperTrait.php
@@ -318,7 +318,7 @@ trait PaginationHelperTrait
         $query_params = [
                 'expand' => 'false',
             ] + $query_params;
-        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT == $this->getClient()->getEndpoint());
+        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT === $this->getClient()->getEndpoint());
         if ($pager) {
             return $this->getResultsInRange($pager, $query_params, $expandCompatibility);
         } else {
@@ -366,7 +366,7 @@ trait PaginationHelperTrait
 
         $uri = $this->getBaseEndpointUri()->withQuery(http_build_query($query_params));
         $response = $this->getClient()->get($uri);
-        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT == $this->getClient()->getEndpoint());
+        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT === $this->getClient()->getEndpoint());
 
         $ids = $this->responseToArray($response, $expandCompatibility);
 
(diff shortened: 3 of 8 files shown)


src/Utility/ResponseToArrayHelper.php
```diff
@@ -34,18 +34,36 @@ trait ResponseToArrayHelper
...
             try {
-                return (array) $this->jsonDecoder()->decode((string) $response->getBody(), 'json');
+                $decoded = (array) $this->jsonDecoder()->decode((string) $response->getBody(), 'json');
+
+                if ($expandCompatibility) {
+                    $root = reset($decoded);
+                    $decoded = [];
+                    foreach ($root as $item) {
+                        $item = (array) $item;
+                        $decoded[] = reset($item);
```
cnovak: Same question: Is it common to use `reset` to get the first array value versus using an index?

src/Utility/ResponseToArrayHelper.php
```diff
@@ -34,18 +34,36 @@ trait ResponseToArrayHelper
...
+    protected function responseToArray(ResponseInterface $response, bool $expandCompatibility = false): array
     {
         if ($response->getHeaderLine('Content-Type') &&
             0 === strpos($response->getHeaderLine('Content-Type'), 'application/json')) {
             try {
-                return (array) $this->jsonDecoder()->decode((string) $response->getBody(), 'json');
+                $decoded = (array) $this->jsonDecoder()->decode((string) $response->getBody(), 'json');
+
+                if ($expandCompatibility) {
+                    $root = reset($decoded);
```
cnovak: Is it common to use reset to get the first array value versus using an index? Just wondering about readability. If so, maybe comments would help explain this section.

cnovak (changes requested):
TryingThe `NonPaginatedEntityIdListingControllerTrait` class also sends the `expand=false` parameter, are we sure we do not need to modify it?

src/Controller/PaginationHelperTrait.php
```diff
@@ -359,8 +366,9 @@ private function listEntityIdsWithoutCps(PagerInterface $pager = null, array $qu
 
         $uri = $this->getBaseEndpointUri()->withQuery(http_build_query($query_params));
         $response = $this->getClient()->get($uri);
+        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT == $this->getClient()->getEndpoint());
```
mxr576: Strict comparison

src/Controller/NonPaginatedEntityListingControllerTrait.php
```diff
@@ -50,7 +50,7 @@ public function getEntities(): array
     /**
      * @inheritdoc
      */
-    abstract protected function responseToArray(ResponseInterface $response): array;
+    abstract protected function responseToArray(ResponseInterface $response, bool $expandCompatibility = false): array;
```
mxr576, line 53: Wow, this is a major change... why a response to array converter have to know about whether the org supports expand or not? I have to think about how this could be solved in a better way if it could be.

mxr576 (changes requested):

src/Controller/NonPaginatedEntityIdListingControllerTrait.php
```diff
@@ -40,12 +41,13 @@ public function getEntityIds(): array
         ];
         $uri = $this->getBaseEndpointUri()->withQuery(http_build_query($query_params));
         $response = $this->getClient()->get($uri);
+        $expandCompatibility = (ClientInterface::HYBRID_ENDPOINT == $this->getClient()->getEndpoint());
```
mxr576: What if hybrid gets supported on on-premise installations as well? I mean, should we check rather the hybrid feature enabled on the org (`Apigee\Edge\Utility\OrganizationFeatures::isHybridEnabled()`) and loading the org for that, just like we do to[ figure out whether CPS is supported or not](https://github.com/apigee/apigee-client-php/blob/2.x-hybrid/src/Controller/PaginationHelperTrait.php#L134-L136), instead of relying on the used endpoint?  We should be consistent in how we are identifying whether we are connected to a hybrid org or not. We should not use different approaches in different places.

src/Controller/NonPaginatedEntityListingControllerTrait.php
mxr576: Maybe we should rather introduce a `responseToArrayHyrid()` method and call that instead of adding too much complexity to the existing method. Where `responseToArray()` is called you already have more context (see above) so you can decide which method should be called.

mxr576: Btw, the returned structure is kinda similar now to the structure that Apigee Edge returns when `expand=true`, the major difference is the nested items are not objects just ids. See `\Apigee\Edge\Controller\PaginationHelperTrait::listEntitiesWithoutCps()` for example, where `responseToArray()` first parse the complete JSON to an array, then we drop the entity type-specific key ("proxies" in this example) and then iterate over the second level array items

```php
[
    {
      "name": "helloworld"
    },
    {
      "name": "weather"
    }
  ]
```

to parse them to entity objects. In this case. we would not parse them entity objects, we would only return the nested values with an `array_reduce()` call.

src/Utility/ResponseToArrayHelper.php
arlina-espinoza: It is not uncommon, and in this case, it's simpler because we don't know the index of the first item (because it's an associative array and the key of the first item might be different depending on which entity endpoint was requested).

src/Controller/NonPaginatedEntityIdListingControllerTrait.php
arlina-espinoza: Hybrid's management API will always be called through the Apigee-hosted management plane (https://docs.apigee.com/hybrid/beta2/what-is-hybrid#api-programs-in-a-hybridmulti-cloud-world), so we can be sure this will not change. 
As for checking if the organization `isHybridEnabled()` or not, I agree that it would make for a more consistent use, however, it will:
- Add another call to the API when we can already deduce this from the client.
- Require that some controllers use `OrganizationControllerAwareTrait` and `OrganizationAwareControllerTrait`. The current pattern of using these traits is to make the controllers have an `$organizationController` and `$organization` properties, which can be injected through constructors. Changing constructors would then be a breaking change, because even if we can set a default value for `$organizationController`, we can't do that for `$organization`.
So overall I still think checking the client endpoint works for this case. What do you think?

src/Controller/NonPaginatedEntityListingControllerTrait.php
arlina-espinoza: I added a helper method `normalizeExpandFalseForHybrid()` to separate this logic further.

src/Controller/PaginationHelperTrait.php
arlina-espinoza: Fixed.

src/Utility/ResponseToArrayHelper.php
arlina-espinoza: Added helper method `normalizeExpandFalseForHybrid()` with comments.

src/Controller/NonPaginatedEntityListingControllerTrait.php
cnovak: I am good w/the approach of having a expandCompatibility param.

cnovak (approved):

src/Controller/NonPaginatedEntityListingControllerTrait.php
arlina-espinoza: > Maybe we should rather introduce a `responseToArrayHyrid()` method and call that instead of adding too much complexity to the existing method. Where `responseToArray()` is called you already have more context (see above) so you can decide which method should be called.

Btw, I considered it, but the "appropriate context" to figure out whether to use `responseToArray()` or `responseToArrayHybrid()` might end up being some spaguetti code, as it might be several calls above - it's not just where `responseToArray()` is called, but where we the query param expand=false is used, so I feel the current approach is more practical.

PR #88, opened 2019-11-08
[#86] Encode parameters in entity endpoint URIs.

### LINKED ISSUE #86: Hybrid API endpoint parameters should be URL encoded
A request to the Hybrid API of the type `/organizations/{{organizationsId}}/developers/{{developer_email}}` will return a 404 if the developer email contains an invalid character such as "+". Hybrid API requires such parameters to be URL encoded.

Example of failing request: `/organizations/{{organizationsId}}/developers/user+1@example.com`
Example of correct request: `/organizations/{{organizationsId}}/developers/user%2B1%40example.com`

arlina-espinoza:
Fixes #86 by URL encoding the entity IDs before adding them to the URI.

### DIFF
--- a/src/Api/Management/Controller/AppCredentialController.php
+++ b/src/Api/Management/Controller/AppCredentialController.php
@@ -124,6 +124,7 @@ abstract class AppCredentialController extends EntityController implements AppCr
      */
     public function setApiProductStatus(string $consumerKey, string $apiProduct, string $status): void
     {
+        $apiProduct = rawurlencode($apiProduct);
         $uri = $this->getBaseEndpointUri()
             ->withPath("{$this->getBaseEndpointUri()}/keys/{$consumerKey}/apiproducts/{$apiProduct}")
             ->withQuery(http_build_query(['action' => $status]));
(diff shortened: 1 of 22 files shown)


cnovak (approved):
Approving, but maybe a unit test would be good here to make sure any changes are still valid in the future, especially testing against the actual endpoint for errors if encoding changes. Feel free to add as issue also.

mxr576 (changes requested):
IIRC only a developer's email address could contain problematic characters in the path because other entitys' have autogenerated machine names or UUIDs. So I would only try to address this in the developer controller.

mxr576:
I would also check whether there are other possible parameters in the paths that are not machine names or UUIDs so they could contain invalid characters,  because this only fixes the entity id parameter is a certain place.

mxr576:
> Approving, but maybe a unit test would be good here to make sure any changes are still valid

Agree, and I think we should not pre-approve PRs that require additional test coverage. We had ~90% code coverage when 2.0.0 released, we should keep the test coverage. https://codecov.io/gh/apigee/apigee-client-php

arlina-espinoza:
> IIRC only a developer's email address could contain problematic characters in the path because other entitys' have autogenerated machine names or UUIDs. So I would only try to address this in the developer controller.

@mxr576   Looking at the docs and testing, other entities might have problematic IDs. Developers can have `+` on their emails, and other entity machine names may contain spaces (apps, API products, attributes, see https://docs.apigee.com/api-platform/reference/naming-guidelines for some examples, although there are more). 
So I combed through the code searching for other places where url encoding might be needed , and added the encoding in multiple endpoints. I also replaced `urlencode()` with `rawurlencode()` because the API requires spaces to be encoded as `%20` instead of a `+`.

Note: organizations, environments, companies and app keys cannot contain spaces in their machine names, either because they are autogenerated or have some validation that prevents that. 

@cnovak I also added a trait to test the saving and loading of entities with machine names that use URL encoding, so it would be helpful to give this PR another review, please.

cnovak:
LGTM now, thanks for adding test

PR #91, opened 2019-11-18
Verify empty response on normalizeExpandFalseForHybrid().

arlina-espinoza:
Fixing an issue related to #85 when the response is empty. Also added a test scenario for this.

### DIFF
--- a/src/Utility/ResponseToArrayHelper.php
+++ b/src/Utility/ResponseToArrayHelper.php
@@ -89,7 +89,7 @@ trait ResponseToArrayHelper
     {
         // If empty, no further processing is needed.
         if (empty($responseArray)) {
-            return  $responseArray;
+            return $responseArray;
         }
 
         // Ignore entity type key from response, ex.: apiProduct.
--- a/src/Utility/ResponseToArrayHelper.php
+++ b/src/Utility/ResponseToArrayHelper.php
@@ -87,6 +87,11 @@ trait ResponseToArrayHelper
      */
     protected function normalizeExpandFalseForHybrid(array $responseArray): array
     {
+        // If empty, no further processing is needed.
+        if (empty($responseArray)) {
+            return  $responseArray;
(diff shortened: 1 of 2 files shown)


src/Utility/ResponseToArrayHelper.php
```diff
@@ -87,6 +87,11 @@ protected function responseToArray(ResponseInterface $response, bool $expandComp
      */
     protected function normalizeExpandFalseForHybrid(array $responseArray): array
     {
+        // If empty, no further processing is needed.
+        if (empty($responseArray)) {
+            return  $responseArray;
```
minnur: an extra space between return and the variable.

tests/Utility/ResponseToArrayHelperTest.php
```diff
@@ -42,21 +42,39 @@ protected function setUp(): void
...
         $this->assertEquals($decodedEdgeStyle, $decodedWithCompatibility);
     }
+
+    /**
+     * DataProvider for testExpandCompatibility.
+     *
+     * @return array
+     *   The array of values to run tests on.
+     */
+    public function expandCompatibilityDataProvider() {
```
minnur: Maybe this should be an array and converted to json via json_encode() so it is simpler to make changes ?

minnur (changes requested):

src/Utility/ResponseToArrayHelper.php
arlina-espinoza: Fixed!

tests/Utility/ResponseToArrayHelperTest.php
arlina-espinoza: Yeah, it would be clearer what the format of Hybrid responses are. Pushed the change. :+1:

minnur (approved):
Looks good to me.

PR #92, opened 2019-11-19
[#296-apigee edge drupal]

arlina-espinoza:
Helps fix https://github.com/apigee/apigee-edge-drupal/issues/296. Two issues are resolved:
- Retrieve the "Edge error code" correctly (ie, the error code of type `developer.service.AppDoesNotExist`), as Hybrid returns it under a substructure, which we try to parse.
- Hybrid is strict about headers, and it does not accept `$headers['Content-Type'] = 'application/json; charset=utf-8';` - it returns an error 400 _(Request contains an invalid argument)_ for POST/PUT. It will only use `$headers['Content-Type'] = 'application/json`. Current Edge API also supports this change.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -208,7 +208,7 @@ class Client implements ClientInterface
     public function post($uri, $body = null, array $headers = []): ResponseInterface
     {
         if (!isset($headers['Content-Type'])) {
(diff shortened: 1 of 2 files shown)


minnur (approved):
Looks good to me.

PR #100, opened 2020-04-01
Add support to phpdocumentor/reflection-docblock:^5.0

### LINKED ISSUE #99: Add/check compatibility with phpdocumentor/reflection-docblock:^5.0
`phpdocumentor/reflection-docblock:5.1.0` gets installed by default when you spin up a new Drupal test environment from https://github.com/drupal-composer/drupal-project and this library currently only supports `"phpdocumentor/reflection-docblock": "^3.0 || ^4.0",`  so the installed version has to be dowgraded to be able to install the https://github.com/apigee/apigee-edge-drupal and this library as its dependency. 

Suggested solution:
Validate compatibility with the 5.x version and it is supported, add it to the compatibility this.

cnovak:
Fixes #99, upgrade to phpdocumentor/reflection-docblock ^5.0, tests pass. The library `symfony/property-info` is the reason we need `phpdocumentor/reflection-docblock`.

However, it looks like [`symfony/property-info` is still testing against 4.x versions](https://github.com/symfony/property-info/blob/master/composer.json#L34) so I am not sure if we want to upgrade yet. 

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ You can find more information about this in the [related pull request](https://g
 * Setters on entities now accepts variable lengths arguments instead of an array. This way we can leverage PHP's built-in type check on these methods as well.
 * Developer and company entities extends and implements one new parent class and interface: AppOwner and AppOwnerInterface.
 * Refactored management API tests.
-  * New environment variable: APIGEE_EDGE_PHP_CLIENT_API_CLIENT
+* New environment variable: APIGEE_EDGE_PHP_CLIENT_API_CLIENT
 * Bumped minimum required versions from php-client/httplug and php-client/client-common packages.
 * Travis CI: Removed PHP nightly builds from the test matrix, fixed failed tests caused by Composer process timeout.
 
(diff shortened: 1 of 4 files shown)


shadcn (approved):
Test. This looks good. LGTM. Thanks @cnovak

mxr576 (dismissed):
> However, it looks like symfony/property-info is still testing against 4.x versions so I am not sure if we want to upgrade yet.

Good catch Chris! I would also try to add support to 5.x versions of all required Symfony packages by the lib. Because we have proper highest lowest testing in this package if something gets broken with 5.x packages, automated tests should discover that. 

I see you also added other documentation changes to this PR, I would create a separate PR for those to keep a clearer Git history and commit could be reverted more easily if needed.

mxr576:
It is more and more painful installing the Apigee Edge module on a Drupal project which uses `drupal/core-dev` because of this... can we give a higher priority this PR? As of today, this simple trick is required to install the module:
`composer require drupal/apigee_edge --no-update && composer update drupal/apigee_edge drupal/coder phpdocumentor/reflection-docblock phpspec/prophecy --with-dependencies`

arlina-espinoza:
@googlebot I consent.

arlina-espinoza:
Rebased against current 2.x and extracted documentation changes to #109 .

arlina-espinoza:
> > However, it looks like symfony/property-info is still testing against 4.x versions so I am not sure if we want to upgrade yet.
> 
> Good catch Chris! I would also try to add support to 5.x versions of all required Symfony packages by the lib. Because we have proper highest lowest testing in this package if something gets broken with 5.x packages, automated tests should discover that.

Drupal core is still using Symfony 4.4 and there are no plans to upgrade until Drupal 10 (and then probably to Symfony 6): https://www.drupal.org/project/drupal/issues/3161889

I'll create a follow up task for that, though (https://github.com/apigee/apigee-client-php/issues/110).

PR #101, opened 2020-05-15
Use toDate instead of endDate for revenue report criteria

shadcn:
See https://apidocs.apigee.com/monetize/apis/post/organizations/%7Borg_name%7D/%7Breport_type%7D

Existing tests are using `toDate` as well: https://github.com/apigee/apigee-client-php/blob/02dd82ec6bcb6d4f4ffca12f072c4ac6b992d998/tests/Api/Monetization/Denormalizer/ReportCriteriaDenormalizerTest.php#L110


arlina-espinoza (approved):
LGTM. The monetization docs make no mention in the report criteria about an `endDate`, just a `toDate`, so this is a bug fix and should not break anything (https://docs.apigee.com/api-platform/monetization/create-reports#critconfigapi).

PR #106, opened 2020-10-12
Drop support of HTTPlug 1.x

### LINKED ISSUE #41: Support PSR-17 and PSR-18 interfaces and php-http 2.x packages
## Preface
First [PSR-17](https://www.reddit.com/r/PHP/comments/93di46/psr17_http_factory_has_been_approved_and_is_now/) and later [PSR-18](https://www.reddit.com/r/PHP/comments/9smt69/psr18_http_client_has_been_accepted/) accepted and PHP-HTTP libraries has started to update their code to rely on new interfaces and add return types to their methods. New major versions got released from these libraries because these changes caused BC breaks with the previous versions.

## Problem
Libraries that we use:
* php-http/httplug: https://github.com/php-http/httplug/compare/v1.1.0...v2.0.0
   * We hare already used these return types so these are no BC breaking changes for this library.
* php-http/client-common: https://github.com/php-http/client-common/compare/1.9.0...2.0.0-RC1 (only RC1 is available at this moment)
  * Good news, this upcoming version is finally contains a proper fix for https://github.com/php-http/client-common/pull/113 (implemented in https://github.com/php-http/client-common/pull/141). Bad news, this fix is only available in the 2.x branch and [it can not be back-ported](https://github.com/php-http/client-common/issues/115#issuecomment-456373787) officially to the 1.x branch. We could back-port this patch for this library but applying it would be only required if < 2.0.0 version is installed from php-http/client-common. Cweagans' composer patches (that we use and recommend) does not support version based patch restrictions (only [vaimo's version](https://github.com/vaimo/composer-patches#patches-version-restriction) has that feature) therefore applying this patch could not be automated. Patching the library would cause extra complexity for consumers. Consequently, it would be better to set the minimum requirement to the php-http/client-common:^2.0.
  * For the first sight, this new major release only requires some [small changes](https://github.com/mxr576/apigee-client-php/commit/502181c98cab1330d4e45d6266df14390dc172da) in this library. Although, if you check these changes more closely then you can see they makes impossible to support both the 1.x version and the 2.x version from the library in the same time.
    * Different exception classes are being used in the Journal and History classes in the 1.x and 2.x version: https://github.com/php-http/client-common/pull/163.
       * This could be fixed by not extending the Journal interface from the client-common in this library's JournalInterface interface and creating our own History plugin that depends on this library's JournalInterface and not client common's definition. That would cause an additional maintenance cost for us.
      * Or by implementing something like Nyholm suggeted: https://github.com/php-http/client-common/issues/115#issuecomment-456355427
* php-http/guzzle6-adapter: https://github.com/php-http/guzzle6-adapter/compare/v1.1.1...v2.0.1 This is only a dev dependency so it should not be a blocker, just added to the list for the sake of completeness.

Related conversion which contains some info about the maintenance and support of the 1.x (older) php-http packages and more: https://github.com/php-http/client-common/issues/115#issuecomment-456012112

## Proposed solution

Because it seems there is no good way to support both 1.x and 2.x version from all required php-http libraries in the current 2.x branch therefore we should release a new major from this library. The new 3.x version should rely on PSR-17 and PSR-18 interfaces instead of interfaces provided by php-http/httplug.

> New client implementations and consumers should use the PSR-18 interfaces directly. In the long term, we expect PSR-18 to completely replace the need for HTTPlug.

(From php-http/httplug's README.md)

The new 3.x version should support only 2.x versions from php-http/client-common and php-http/guzzle6-adapter (as dev dependency) libraries.

In addition, it seems we should wait for the release of the 1.6.0 from the php-http/discovery and depend on that: https://github.com/php-http/discovery/issues/134.

The same resolution happens in case of the Mailgun library: https://github.com/mailgun/mailgun-php/pull/522

## Questions

1. What should happen with the 2.x version? Should we abandon it completely and force developers to update to the 3.x version or mark it with a similar status like Drupal's "[Maintenance fixes only](https://www.drupal.org/node/1066982)". ? If we choose the second option, how long we should wait until we mark it as unsupported? 

(This is a follow up on #24.)

mxr576:
Closes #41


composer.json
```diff
@@ -20,16 +20,16 @@
         "fightbulc/moment": "^1.26",
         "firebase/php-jwt": "^5.0",
         "league/period": "^3.4",
-        "php-http/client-common": "^1.9",
+        "php-http/client-common": "^2.0",
         "php-http/client-implementation": "^1.0",
-        "php-http/discovery": "^1.0",
-        "php-http/httplug": "^1.1",
+        "php-http/discovery": "^1.6.0",
```
mxr576, line 25: Because: https://github.com/php-http/discovery/blob/1.6.0/CHANGELOG.md#160---2019-01-23

composer.json
```diff
@@ -20,16 +20,16 @@
...
+        "php-http/httplug": "^2.0",
         "php-http/message": "^1.0",
         "php-http/message-factory": "^1.0",
         "phpdocumentor/reflection-docblock": "^3.0 || ^4.0",
         "psr/http-message": "^1.0",
         "symfony/options-resolver": "^3.4 || ^4.0",
-        "symfony/property-access": "^3.4 || ^4.0",
+        "symfony/property-access": "^3.4.41 || ^4.4.9",
```
mxr576, line 32: Because https://github.com/symfony/property-access/commit/e1a6c91c0007e45bc1beba929c76548ca9fe8a85

mxr576:
@arshad @arlina-espinoza @cnovak hi, may I get a review on this? These are old dependencies and having them here blocks upstream projects to depend on newer, supported versions if they would like to use them together with this library.

arlina-espinoza (changes requested):
@mxr576 When trying to install with the latest template for Kickstart (https://github.com/apigee/devportal-kickstart-project-composer/blob/8.x/composer.json), I get the following errors:

```console
arlina@system76:/v/w/h/c/d9$ composer require apigee/apigee-client-php:"dev-issue/41 as 2.0.6" --update-with-dependencies
./composer.json has been updated
Loading composer repositories with package information                                              Updating dependencies (including require-dev)         
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Conclusion: don't install apigee/apigee-client-php 2.0.6|remove php-http/guzzle6-adapter v1.1.1
    - Conclusion: install apigee/apigee-client-php 2.0.6
    - Conclusion: don't install php-http/httplug 2.2.0
    - apigee/apigee-client-php 2.0.6 requires php-http/client-implementation ^1.0 -> satisfiable by php-http/guzzle6-adapter[v1.1.1].
    - Conclusion: don't install php-http/guzzle6-adapter v1.1.1
    - Conclusion: remove php-http/httplug v1.1.0
    - Conclusion: don't install php-http/httplug 2.1.0
    - Installation request for apigee/apigee-client-php dev-issue/41 as 2.0.6 -> satisfiable by apigee/apigee-client-php[dev-issue/41].
    - Installation request for drupal/apigee_m10n (locked at 1.7.0, required as ^1.0.0) -> satisfiable by drupal/apigee_m10n[1.7.0].
    - php-http/guzzle6-adapter v1.1.1 requires php-http/httplug ^1.0 -> satisfiable by php-http/httplug[v1.1.0, v1.0.0, v1.0.0-RC1, v1.0.0-alpha, v1.0.0-alpha2, v1.0.0-alpha3, v1.0.0-beta].
    - Can only install one of: php-http/httplug[v1.1.0, 2.x-dev].
    - Can only install one of: php-http/httplug[2.x-dev, v1.1.0].
    - Can only install one of: php-http/httplug[v1.0.0, 2.x-dev].
    - Can only install one of: php-http/httplug[v1.0.0-RC1, 2.x-dev].
    - Can only install one of: php-http/httplug[v1.0.0-alpha, 2.x-dev].
    - Can only install one of: php-http/httplug[v1.0.0-alpha2, 2.x-dev].
    - Can only install one of: php-http/httplug[v1.0.0-alpha3, 2.x-dev].
    - Can only install one of: php-http/httplug[v1.0.0-beta, 2.x-dev].
    - apigee/apigee-client-php 2.0.6 requires php-http/httplug ^2.0 -> satisfiable by php-http/httplug[2.1.0, 2.2.0, 2.x-dev, v2.0.0].
    - Conclusion: don't install php-http/httplug v2.0.0


Installation failed, reverting ./composer.json to its original content.
```

Looking into what other dependencies are locking php-http/httplug to ^1.0, I find:

```console
arlina@system76:/v/w/h/c/d9$ composer why php-http/httplug -t
php-http/httplug v1.1.0 HTTPlug, the HTTP client abstraction for PHP
├──apigee/apigee-client-php 2.0.5 (requires php-http/httplug ^1.1)
│  ├──apigee/devportal-kickstart-project (requires apigee/apigee-client-php dev-issue/41 as 2.0.6)
│  ├──drupal/apigee_edge 1.18.0 (requires apigee/apigee-client-php ^2.0.4)
│  │  ├──apigee/apigee_devportal_kickstart 1.21.0 (requires drupal/apigee_edge ^1.16)
│  │  │  └──apigee/devportal-kickstart-project (requires apigee/apigee_devportal_kickstart ^1.0.0)
│  │  └──drupal/apigee_m10n 1.7.0 (requires drupal/apigee_edge ^1.16)
│  │     └──apigee/devportal-kickstart-project (requires drupal/apigee_m10n ^1.0.0)
│  └──drupal/apigee_m10n 1.7.0 (requires apigee/apigee-client-php ~2.0.3)
│     └──apigee/devportal-kickstart-project (requires drupal/apigee_m10n ^1.0.0)
├──php-http/client-common 1.10.0 (requires php-http/httplug ^1.1)
│  └──apigee/apigee-client-php 2.0.5 (requires php-http/client-common ^1.9)
│     ├──apigee/devportal-kickstart-project (requires apigee/apigee-client-php dev-issue/41 as 2.0.6)
│     ├──drupal/apigee_edge 1.18.0 (requires apigee/apigee-client-php ^2.0.4)
│     │  ├──apigee/apigee_devportal_kickstart 1.21.0 (requires drupal/apigee_edge ^1.16)
│     │  │  └──apigee/devportal-kickstart-project (requires apigee/apigee_devportal_kickstart ^1.0.0)
│     │  └──drupal/apigee_m10n 1.7.0 (requires drupal/apigee_edge ^1.16)
│     │     └──apigee/devportal-kickstart-project (requires drupal/apigee_m10n ^1.0.0)
│     └──drupal/apigee_m10n 1.7.0 (requires apigee/apigee-client-php ~2.0.3)
│        └──apigee/devportal-kickstart-project (requires drupal/apigee_m10n ^1.0.0)
└──php-http/guzzle6-adapter v1.1.1 (requires php-http/httplug ^1.0)
   └──drupal/apigee_edge 1.18.0 (requires php-http/guzzle6-adapter ^1.1.1)
      ├──apigee/apigee_devportal_kickstart 1.21.0 (requires drupal/apigee_edge ^1.16)
      │  └──apigee/devportal-kickstart-project (requires apigee/apigee_devportal_kickstart ^1.0.0)
      └──drupal/apigee_m10n 1.7.0 (requires drupal/apigee_edge ^1.16)
         └──apigee/devportal-kickstart-project (requires drupal/apigee_m10n ^1.0.0)

```

So it seems that to drop support for HTTPlug 1.x we would also need to require the 2.x from the `apigee_edge` module. Does this sound right to you?

mxr576:
@arlina-espinoza yes it does, just `php-http/guzzle6-adapter v1.1.1` has to be bumped to ^2.0, that is all.

arlina-espinoza (approved):
LGTM, created https://github.com/apigee/apigee-edge-drupal/pull/525 to support it from the `drupal/apigee_edge` side.

Issue #87, opened 2019-11-08
Hybrid API: Call to environments listing does not support expand=false


arlina-espinoza:
On the Hybrid API: Making a call to `/organizations/***organization***/environments?expand=false`
fails with error 400. It says the parameter "expand" isn't supported for this resource. 
Full response:

```
TTP/1.1 400 Bad Request
Vary: X-Origin, Referer, Origin,Accept-Encoding
Content-Type: application/json; charset=UTF-8
Date: Fri, 08 Nov 2019 20:12:52 GMT
Server: ESF
Cache-Control: private
X-XSS-Protection: 0
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Alt-Svc: quic=":443"; ma=2592000; v="46,43",h3-Q050=":443"; ma=2592000,h3-Q049=":443"; ma=2592000,h3-Q048=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000
Accept-Ranges: none
Transfer-Encoding: chunked

{
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"expand\": Cannot bind query parameter. Field 'expand' could not be found in request message.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "description": "Invalid JSON payload received. Unknown name \"expand\": Cannot bind query parameter. Field 'expand' could not be found in request message."
          }
        ]
      }
    ]
  }
}

```

arlina-espinoza:
Fixed by #90 .

PR #114, opened 2020-12-17
Apigee hybrid /organizations call does not allow expand=true.

giteshk:
Fixed how organizations are fetched for Apigee Hybrid.

### DIFF
--- a/src/Api/Management/Controller/OrganizationController.php
+++ b/src/Api/Management/Controller/OrganizationController.php
@@ -22,10 +22,10 @@ use Apigee\Edge\Api\Management\Entity\Organization;
 use Apigee\Edge\Api\Management\Serializer\OrganizationSerializer;
 use Apigee\Edge\ClientInterface;
 use Apigee\Edge\Controller\AbstractEntityController;
+use Apigee\Edge\Controller\BaseEndpointAwareControllerTrait;
+use Apigee\Edge\Controller\ClientAwareControllerTrait;
 use Apigee\Edge\Controller\EntityCrudOperationsControllerTrait;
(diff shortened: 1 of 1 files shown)


src/Api/Management/Controller/OrganizationController.php
```diff
@@ -51,6 +51,28 @@ public function __construct(ClientInterface $client, ?EntitySerializerInterface
...
+        if (ClientInterface::HYBRID_ENDPOINT == $this->client->getEndpoint()) {
+            $entities = [];
+            foreach ($responseArray as $item) {
+                $entities[$item->organization] = $this->load($item->organization);
+            }
+
+            return $entities;
+        }
+
+        return $this->responseArrayToArrayOfEntities($responseArray);
```
mxr576, line 73: Can't we just use the following logic instead?

```
if (ClientInterface::HYBRID_ENDPOINT == $this->client->getEndpoint()) {
  // Do the custom stuff and return
}

return parent::getEntities(); // parent is the inherited method from the trait.

```

You may need to alias the inherited trait methods for this.

mxr576 (changes requested):

mxr576:
>Apigee hybrid ... does not allow expand=true

only here or on other API (entity) endpoints are also affected? IIRC this is not the first Hybrid workaround for `expand=true`. If `expand` is not working in Hybrid as it works in public/private cloud then it should be addressed globally not just by hotfixing endpoints one by one. 

See:
#87 
#91 
#85

cnovak:
Thanks much for your excellent code review @mxr576 . It looks like this should be changed to follow the patterns that @mxr576 pointed out.

PR #119, opened 2021-01-18
DX: Make app credential generation safer

mxr576:
Deprecate possibility of overriding app attributes and callback urls
when a new app credential is generated. (Even if the API allows it.)

Admirably, this change needs test coverage.


mxr576:
I am not sure what has changed on the Apigee API backend but it looks like sending _app's_ callbackurl and attributes is again, mandatory, otherwise it is lost. I had automated tests that made sure the patch was worked as expected, but now it fails and not because a code related change on the PHP side.

PR #120, opened 2021-01-20
[TBD] OAuth: URL encoded credentials are not parsed correctly by the backend

mxr576:
If a value contains any special character (like % in a password) then it gets encoded which leads to a malformed payload.

Needs test coverage.

### DIFF
--- a/src/HttpClient/Plugin/Authentication/Oauth.php
+++ b/src/HttpClient/Plugin/Authentication/Oauth.php
@@ -116,25 +116,28 @@ class Oauth extends AbstractOauth
     {
         if ($this->tokenStorage->getRefreshToken()) {
             $body = [
-                'grant_type' => 'refresh_token',
-                'refresh_token' => $this->tokenStorage->getRefreshToken(),
+                'grant_type=refresh_token',
+                "refresh_token={$this->tokenStorage->getRefreshToken()}",
             ];
         } else {
             $body = [
-                'grant_type' => 'password',
-                'username' => $this->username,
-                'password' => $this->password,
+                'grant_type=password',
+                "username={$this->username}",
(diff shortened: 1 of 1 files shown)


mxr576:
Something is off here, the original code supposes to be correct, but if it is correct then special characters should not be allowed in values (and keys) because `application/x-www-form-urlencoded` [means](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) that "Non-alphanumeric characters in both keys and values are percent encoded:". Consequently, for example, if a value has `%` then it has to be encoded to `%25` , but that malformed the original input and authentication fails.

I checked what the `get_token` [script](https://docs.apigee.com/api-platform/system-administration/auth-tools) does and it also uses curl's `--data-urlencode` parameter (instead of just `--data`) for username and password, which has the same unexpected side-effect.

Although, there is this other documentation about [Get the machine user's OAuth2 tokens](https://docs.apigee.com/api-platform/system-administration/automating-tasks#get-the-machine-users-oauth2-tokens) and there is a "cheat" in that because it does not use either `--data` or `--data-urlencode`, it just builds the raw request body (without encoding the values).

@cnovak can you share your thoughts about it, or check with engineering what is the expected behavior here?

cnovak:
Thanks @mxr576 for bringing this to our attention, will have the team look into it.

mxr576:
Updated the title, it reflects the issue more clearly based on my current knowledge.
Yesterday I ran an additional test with Postman and the same issue happened. When I set the `application/x-www-form-urlencoded` header, Postman automatically started to percent encode the key-value pairs in the request body which lead to the same issue.

mxr576:
`sudo re-open`

PR #121, opened 2021-01-27
restrict polyfill-mbstring to 1.20.0 / Fix travis-ci

phdhiren:
With latest symfony/polyfill-mbstring, introduced PHP 8, which is syntax wise failing for PHP 7.1 or PHP 7.2
Till the time the module's PHP 8 combability and/or other multiple module updates restricting polyfill-mbstring library to beyond 1.20.0 so the test on travis-ci can pass.


mxr576:
> Till the time the module's PHP 8 combability and/or other multiple module updates restricting polyfill-mbstring library to beyond 1.20.0 so the test on travis-ci can pass.

Module? Are you referring to the Apigee Edge module? Because this library does not depend directly on this polyfill, are you saying it depends on this lib indirectly?

phdhiren:
I mean this library - thanks for correcting.
If you've noticed with recent tests is failing due to PHP 8 code in symfony/polyfill-mbstring library which is second level dependency of friendsofphp/php-cs-fixer

After their commit - [Adjust mbstring polyfill for PHP 8](https://github.com/symfony/polyfill-mbstring/commit/11d0d87a1d1ef6a3d8158fcb756387786490cd08) CircleCI is failing because of PHP syntax failing below PHP7.4


```
ERROR: ParseError - vendor/symfony/polyfill-mbstring/bootstrap80.php:125:86 - Syntax error, unexpected '=' on line 125
    function mb_scrub(string $string, string $encoding = null): string { $encoding ??= mb_internal_encoding(); return mb_convert_encoding($string, $encoding, $encoding); }

```

So, while the whole library is ready for PHP 8 upgrade would like polyfill-mbstring to restrict to 1.20.0


However, main purpose for the pull to fix CircleCI test, which still didn't worked.

mxr576:
Circle CI? Are you working on replacing Travis here?

phdhiren:
If possible then move to CircleCI else atleast fix Travis

cnovak:
PHP versions we test against and support should be 7.3. and 7.4 since  < 7.3 is end of life: https://www.php.net/supported-versions.php

mxr576:
> PHP versions we test against and support should be 7.3. and 7.4 since < 7.3 is end of life: https://www.php.net/supported-versions.php

Well true, but you cannot be sure about who uses Zend's LTS versions, so throwing way <7.3 is not that simple.  (We have met a customer who does, so 7.1 support is still required...) Also before, if we would like to play nice, before we would throw away support of a PHP version we should announce it in time and let developers time to adapt. Check Symfony's and Drupal's philosophy about this topic. 

https://www.zend.com/sites/zend/files/image/2020-06/image-zend-compare-php-support-zend-server.jpg

mxr576:
> If possible then move to CircleCI else atleast fix Travis

What about Github Actions? It is free for FOSS projects and maybe a better alternative than CircleCI as of today.

cnovak:
@mxr576 this PR is an attempt to fix the failing tests, but was not successful. If you have any ideas on how to fix it would be greatly appreciated.  Since we are having trouble fixing, we decided that we can fix the problem by upgrading PHP version to a supported version (#124). I do agree that a heads up on changes is a great idea, but I am not sure if we can EOL versions.  Any ideas you have on how to get < 7.4 versions working would be appreciated since we need to have running tests in order to make changes to this repo.

cnovak:
> > If possible then move to CircleCI else atleast fix Travis
> 
> What about Github Actions? It is free for FOSS projects and maybe a better alternative than CircleCI as of today.

I have not looked into Github Actions. The only reason I was thinking CircleCi is to make all projects consistent, but it is worth looking into GitHub Actions to @mxr576 's point

gosunilgo:
> > If possible then move to CircleCI else atleast fix Travis
> 
> What about Github Actions? It is free for FOSS projects and maybe a better alternative than CircleCI as of today.

@mxr576 @cnovak  -- Take a look at #127 initial/draft of a ci-pipeline

raakesh-blokhra (approved):
Looks good for fixing current failing issues.

Issue #108, opened 2020-10-29
Notice in recent message logs if description is empty after creating app in Apigee hybrid


varuntit101:
## Description
Notice in recent message logs if description is empty after creating app in Apigee hybrid

## Steps to Reproduce
Create an app with empty description

`Notice: Undefined property: stdClass::$value in Apigee\Edge\Denormalizer\AttributesPropertyDenormalizer->denormalize() (line 57 of \vendor\apigee\apigee-client-php\src\Denormalizer\AttributesPropertyDenormalizer.php)`

mxr576:
@shishir-intelli please add test coverage to #122 if you have time, that is required to merge the fix.

PR #122, opened 2021-01-28
Handle possible null value caused by missing value in response payload

mxr576:
If an attribute's value is an empty string then this API endpoint does not return the "value" in the API response only the "name".

So instead of 

```json
{
    "name": "DisplayName",
    "value": "someapp"
}
```
what is returned is

```json
{
    "name": "DisplayName",
}

```
when

```json
{
    "name": "DisplayName",
    "value": ""
}
```
A related API is: https://apidocs.apigee.com/docs/api-products/1/routes/organizations/%7Borg_name%7D/apiproducts/%7Bapiproduct_name%7D/attributes/%7Battribute_name%7D/get

Needs test coverage.


mxr576:
Fixes #108, which was already reported.

mxr576:
#125 was merged... closing this.

PR #127, opened 2021-02-22
add actions for php-build

gosunilgo:
Initial commit for ci-pipeline with github-actions

- matrix build for multiple php verions
- Noticed symfony property-access test is not available.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -1,0 +1,52 @@
+name: PHP Composer
+
+on:
+  push:
+    branches: [ 2.x ]
+  pull_request:
+    branches: [ 2.x ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    
+    strategy: 
+      fail-fast: false
+      matrix: 
+        php-versions: ['7.1','7.2','7.3', '7.4','8.0']
+        phpunit-versions: ['latest']
+      
+    steps:
+    - uses: actions/checkout@v2
+    - uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php-versions }} #'7.2'
+        tools: composer:v1
+        extensions: mbstring, intl
+        #ini-values: pcov.directory=api #optional, see above for usage.
+        coverage: pcov
+     
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
+    - name: Cache Composer packages
+      id: composer-cache
+      uses: actions/cache@v2
+      with:
+        path: vendor
+        key: ${{ runner.os }}-php-${{ hashFiles('**/composer.json') }}
+        restore-keys: |
+          ${{ runner.os }}-php-
+
+    - name: Install dependencies
+      if: steps.composer-cache.outputs.cache-hit != 'true'
+      run: composer install --prefer-dist --no-progress --no-suggest
+    - name: cache update
+      run: composer update --no-suggest --no-interaction
+      
+
+    - name: Run show
+      run: composer show     
+    - name: Run analyze
+      run: composer test


phdhiren:
@gosunilgo , Thanks for your contribution, we really appreciate it. I took at look at your actions and see there are some errors. Did you expect this to be checked in or are you giving ideas on how it works?
@cnovak , Github Actions looks promising

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
+name: PHP Composer
```
mxr576, line 1: "Test coverage" or sth like that, the workflow file could be also called "testing.yml".

Code style checking and static code analyzer is missing from here, it could go to a separate workflow file or here.

I think this PR should be a "Replace Travis CI" which _completely_ replaces Travis and removes it from the repo. Do you agree?

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches: [ 2.x ]
+  pull_request:
+    branches: [ 2.x ]
```
mxr576, line 7: Scheduled daily builds could be also useful to see if a change in upstream causes breaking change here.

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
mxr576, line 12: Better to depend on a specific version then "latest".

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+     
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
+    - name: Cache Composer packages
+      id: composer-cache
+      uses: actions/cache@v2
+      with:
+        path: vendor
+        key: ${{ runner.os }}-php-${{ hashFiles('**/composer.json') }}
```
mxr576, line 38: The cache key should contain the date because otherwise the same cache entry is being used until there is a change in composer.json. (This is not just a blocker for daily builds, see above)

[This solution](https://github.com/actions/cache/blob/main/examples.md#php---composer) works well when `composer.lock` is being hashed, but libraries should not contain composer.lock and hashing composer.json requires a different solution.

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+        coverage: pcov
+     
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
+    - name: Cache Composer packages
+      id: composer-cache
+      uses: actions/cache@v2
+      with:
+        path: vendor
```
mxr576, line 37: Composer's file cache directory should be cached not the vendor folder, see: 
This solution](https://github.com/actions/cache/blob/main/examples.md#php---composer

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+        restore-keys: |
+          ${{ runner.os }}-php-
+
+    - name: Install dependencies
+      if: steps.composer-cache.outputs.cache-hit != 'true'
+      run: composer install --prefer-dist --no-progress --no-suggest
+    - name: cache update
+      run: composer update --no-suggest --no-interaction
+      
+
```
mxr576, line 48: Nit: unnecessary empty line.

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        path: vendor
+        key: ${{ runner.os }}-php-${{ hashFiles('**/composer.json') }}
+        restore-keys: |
+          ${{ runner.os }}-php-
+
+    - name: Install dependencies
+      if: steps.composer-cache.outputs.cache-hit != 'true'
+      run: composer install --prefer-dist --no-progress --no-suggest
+    - name: cache update
```
mxr576, line 45: This is unnecessary if Composer file cache is being cached.

.github/workflows/php.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    
+    strategy: 
+      fail-fast: false
+      matrix: 
+        php-versions: ['7.1','7.2','7.3', '7.4','8.0']
+        phpunit-versions: ['latest']
```
mxr576, line 18: Unused, the PHPUnit version is defined in composer.json, if that does not support 7.1 - 8.0 then the version range should be extended in composer.json and Composer is going to pick the right version based on the current PHP version.

mxr576 (changes requested):
Great progress!

On the "Checks" tab we should see the workflow in action, can somebody check why it is not running?

.github/workflows/php.yml
mxr576: Highest-lowest testing is also missing from here, see .travis.yml

giteshk:
Thank you @gosunilgo for your contribution. We have recently merged #212 which adds support for GitHub actions.

Issue #115, opened 2020-12-20
Consider handling Hybrid as an entirely new API 


mxr576:
### Problem

I am not sure what are the long term plans for Apigee Hybrid, but as of today, there are known and unknown differences between Apigee APIs on Hybrid and in Public/private cloud. (Is it the right naming convention? 🤔 )

Known differences are described in [this document](https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis) and there are plenty API "families" that are not supported by Hybrid (ex.: Company and Company apps) and there are some unsupported API calls even in the supported ones (ex.: "Get count of API resources" for Developer apps).

Other differences that we have already got bug reports:

1. Data model differences:
  * #65 `createdBy` and `lastModifiedBy` are not available on Hybrid
2. Unsupported API parameters, parameters that work differently
  * `expand=[true,false]` is not working on Hybrid: #85 #87 #114
  * `_optimized` is not supported on Hybrid: #105 #111 

This library can hide complexities from developers, like we did with faking CPS support in #43, to make developers' life easier. Although that workaround can be considered a smaller one compared with those differences that we know about Hybrid.

(Clear difference between #43 and the current code in #114 that in #43 we [trigger a warning](https://github.com/apigee/apigee-client-php/pull/43/files#diff-2fbd1daed02ea1e7408fbe695d74e9bda71ccc4cd82f4e6228571e14f842db50R288), let developers know that a sub-optimal query is being executed by the library (instead of 1 API call, maybe 10 or more is being sent _synchronously_), #114 is currently hiding that fact which can lead to unexpected performance regressions on downstream.)

### Proposed solution

Instead of adding workarounds to our codebase which was designed and built for public/private cloud, introduce a new `Apigee\Edge\Api\Hybrid`(or without `Edge`) or `Apigee\Hybrid\Api` namespace and start collecting those APIs and data models under this namespace that are supported by Hybrid. Ex.: `Apigee\Hybrid\Api\Management\Entity\DeveloperApp`

PR #128, opened 2021-02-25
View purchased API products

divya-intelli:
Listing of purchased API products.

### DIFF
--- a/src/Api/Monetization/Controller/XAcceptedRatePlanController.php
+++ b/src/Api/Monetization/Controller/XAcceptedRatePlanController.php
@@ -1,0 +1,189 @@
+<?php
+
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


src/Api/Monetization/Controller/XAcceptedRatePlanController.php
```diff
@@ -0,0 +1,189 @@
...
+use Apigee\Edge\Api\Monetization\Serializer\XAcceptedRatePlanSerializer;
+use Apigee\Edge\ClientInterface;
+use Apigee\Edge\Controller\EntityListingControllerTrait;
+use Apigee\Edge\Controller\EntityLoadOperationControllerTrait;
+use Apigee\Edge\Serializer\EntitySerializerInterface;
+use DateTimeImmutable;
+use Psr\Http\Message\UriInterface;
+use ReflectionClass;
+
+abstract class XAcceptedRatePlanController extends OrganizationAwareEntityController implements XAcceptedRatePlanControllerInterface
```
mxr576, line 32: Instead of this "X" prefix in class names, should not we leverage namespaces and open a new `Apigee\Edge\Api\ApigeeX` namespace to group implementations that only works on Apigee X? This would also help downstream devs to understand the complexity that they have to deal with.
I think this also relates to #115 somehow

(ps.: `Apigee\Edge\Api\X\... `looks odd to me but also an alternative...)

cnovak: @mxr576 thanks for bringing this up, we will use your suggestion of creating a `Apigee\Edge\Api\ApigeeX` namespace.

raakesh-blokhra:
Closing PR since new PRs created that replace this initial commit.

PR #129, opened 2021-02-25
Added resources and supporting files for getting product-api listing …

shishir-intelli:
Created XProduct entity/Controller and supporting files for getting the listing of products from 5g.

### DIFF
--- a/src/Api/Monetization/Controller/ApiXProductController.php
+++ b/src/Api/Monetization/Controller/ApiXProductController.php
@@ -1,0 +1,92 @@
+<?php
+
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Apigee\Edge\Api\Monetization\Controller;
+
+use Apigee\Edge\Api\Monetization\Entity\ApiXProduct;
+use Apigee\Edge\Api\Monetization\Entity\XRatePlanInterface;
+use Apigee\Edge\Api\Monetization\Serializer\ApiXProductSerializer;
+use Apigee\Edge\ClientInterface;
+use Apigee\Edge\Controller\EntityListingControllerTrait;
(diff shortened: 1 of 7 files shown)


src/Api/Monetization/Controller/ApiXProductController.php
```diff
@@ -0,0 +1,92 @@
...
+use Apigee\Edge\Api\Monetization\Entity\ApiXProduct;
+use Apigee\Edge\Api\Monetization\Entity\XRatePlanInterface;
+use Apigee\Edge\Api\Monetization\Serializer\ApiXProductSerializer;
+use Apigee\Edge\ClientInterface;
+use Apigee\Edge\Controller\EntityListingControllerTrait;
+use Apigee\Edge\Controller\EntityLoadOperationControllerTrait;
+use Apigee\Edge\Serializer\EntitySerializerInterface;
+use Psr\Http\Message\UriInterface;
+
+class ApiXProductController extends OrganizationAwareEntityController implements ApiXProductControllerInterface
```
mxr576, line 30: The naming convention inconsistent with #128. This also a confirmation to me that introducing a dedicated namespace would be better and having "X" in class names _somewhere_ would not be necessary. @cnovak what do you think?

Eg. 

```php
Apigee\Edge\Api\Management\Controller\ApiProductController
Apigee\Edge\Api\Monetization\Controller\ApiProductController
Apigee\Edge\Api\ApigeeX\Monetization\Controller\ApiProductController
Apigee\Edge\Api\ApigeeX\Management\Controller\ApiProductController # is it valid?
```

https://github.com/apigee/apigee-client-php/pull/128#discussion_r583431876

shishir-intelli: @mxr576 ,
For Apigee X, we are working on the same existing m10 module which is 4G, to differentiate 5g files from 4g we have used X in the file names to avoid confusion (_ApiProduct, RatePlan etc names already exists there_).
Yes, we need to be consistent with the names, we will work on this to keep it consistent.
The PR is in a draft as there can be some alterations.

mxr576: Thanks @shishir-intelli , FYI, I had a conversation about this with Chris on an other PR:
https://github.com/apigee/apigee-client-php/pull/128#discussion_r584988198

>  we are working on the same existing m10 module which is 4G, to differentiate 5g files from 4g we have used X in the file names to avoid confusion

My personal opinion about is: This is the PHP library, the fundamental of everything and independent from every dependents. Just because a consumer (like a Drupal module) uses a naming convention, it does not have to be inherited.

shishir-intelli: Yes, I agree with you and we are going ahead with the suggested approach of creating new namespace "Apigee\Edge\Api\ApigeeX".

Thanks @mxr576 .

raakesh-blokhra:
Closing PR - initial commit, discarded,  New PRs created for the same.

PR #133, opened 2021-04-19
App Analytics page - fix PHP error on empty data

### LINKED ISSUE #134: PHP error in case of empty data on Analytics page
## Description
Empty response from the API leads to PHP Errors. 

## Expected Behavior
If empty response is sent back, ignore it instead of throwing an error.

phperror

phdhiren:
Fix PHP error in case of empty data on Analytics page


raakesh-blokhra (approved):
Looks good. Thanks

PR #142, opened 2021-04-30
Fix for m10n #264

divya-intelli:
Fixes m10 #264 It's not clear which rate card applies to which product when a bundle has a rate card per product.

### DIFF
--- a/src/Api/Monetization/Entity/Property/ProductPropertyInterface.php
+++ b/src/Api/Monetization/Entity/Property/ProductPropertyInterface.php
@@ -27,5 +27,5 @@ interface ProductPropertyInterface
      *
      * @return \Apigee\Edge\Api\Monetization\Entity\ApiProduct
      */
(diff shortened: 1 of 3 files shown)


src/Api/Monetization/Entity/Property/ProductPropertyInterface.php
```diff
@@ -0,0 +1,31 @@
...
+use Apigee\Edge\Api\Monetization\Entity\ApiProduct;
+
+interface ProductPropertyInterface
+{
+    /**
+     * Returns the apiproduct information included in rateplan detials.
+     *
+     * @return \Apigee\Edge\Api\Monetization\Entity\ApiProduct
+     */
+    public function getproduct(): ?ApiProduct;
```
phdhiren: It should be camel case

phdhiren (changes requested):

phdhiren (approved):
Changes looks good to me

PR #145, opened 2021-05-14
Added runtimeType and monetizationConfig

phdhiren:
Added runtimeType and monetizationConfig property to check monetization status in Apigee X


divya-intelli (approved):
LGTM

PR #154, opened 2021-06-22
Fix user profile update not working on Apigee X (PUT /developers)

### LINKED ISSUE #153: User update on Apigee X not working
Steps to reproduce:

Try to update First and Last Name on the user on Drupal portal

# Steps to reproduce

1. In the Admin bar, click on "People"
2. Click to edit a user
3. Change a field such as First name or Last Name
4. Click "Save"

**Expected:**
User is updated

**Actual:**
System throws an error in the errorlog. No error is shown on the screen.

```PUT /developers/22222222-8359-49a1-b49b-d5cf959e9d83 HTTP/1.1
>>>>>>>>
PUT /v1/organizations/***organization***/developers/22222222-8359-49a1-b49b-d5cf959e9d83 HTTP/1.1
Host: apigee.googleapis.com
Content-Length: 371
X-Apigee-Edge-Api-Client-Profiler: X-Apigee-Edge-Api-Client-Profiler
Content-Type: application/json
User-Agent: Apigee Edge DevPortal 8.x-1.22 (Apigee Edge PHP Client 2.0.7)
Accept: application/json; charset=utf-8

{"apps":["APIM_Security_Demo_App","apim_mint_demoapp_1","app2"],"attributes":[],"companies":[],"createdAt":1623806970000,"developerId":"22222222-8359-49a1-b49b-d5cf959e9d83","email":"abcd@example.com","firstName":"abcd","lastModifiedAt":1623806970000,"lastName":"abcd","organizationName":"abcd-org1","status":"active","userName":"abcd1"}
<<<<<<<<
HTTP/1.1 400 Bad Request
Vary: X-Origin, Referer, Origin,Accept-Encoding
Content-Type: application/json; charset=UTF-8
Date: Wed, 16 Jun 2021 22:48:12 GMT
Server: ESF
Cache-Control: private
X-XSS-Protection: 0
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Accept-Ranges: none
Transfer-Encoding: chunked

{
    "error": {
        "code": 400,
        "message": "Developer update : resource name 22222222-8359-49a1-b49b-d5cf959e9d83 in the url doesnot match with the name abcd@example.com in the request payload",
        "status": "INVALID_ARGUMENT",
        "details": [
            {
                "@type": "type.googleapis.com\/google.rpc.PreconditionFailure",
                "violations": [
                    {
                        "type": "rest.interceptor.name_mismatch",
                        "subject": "[2002:a17:531:8141::]:4177:vobi17:9855:347726:3196811",
                        "description": "Developer update : resource name 22222222-8359-49a1-b49b-d5cf959e9d83 in the url doesnot match with the name abcd@example.com in the request payload"
                    }
                ]
            },
            {
                "@type": "type.googleapis.com\/google.rpc.RequestInfo",
                "requestId": "3243783479475234489"
            }
        ]
    }
}
Transfer statistics:

total_time: 0.918s
namelookup_time: 0s
connect_time: 0s
pretransfer_time: 0s
starttransfer_time: 0.917s
redirect_time: 0s
appconnect_time: 0s
```

shishir-intelli:
Closes  #153 

Fixed issue by providing the developer `email-ID` instead of the developer `UUID` before clients `EntityUpdateOperationControllerTrait` update function is called.  (`PUT /developers` call)
This works for both 4G/ 5G.

Note: Also for Apigee X and hybrid runtime versions 1.5.0 and 1.5.1, a call to change the developer's email address will not work.

### DIFF
--- a/src/Api/Management/Controller/DeveloperController.php
+++ b/src/Api/Management/Controller/DeveloperController.php
@@ -98,7 +98,7 @@ class DeveloperController extends PaginatedEntityController implements Developer
     /**
      * {@inheritdoc}
      *
-     * We are enforcing email addresses over developer ids (UUIDs) when we are updating user profile.
+     * We are enforcing email addresses over developer-ids (UUIDs) when we are updating user profile.
      *
      * @psalm-suppress UndefinedMethod
      *
--- a/src/Api/Management/Controller/DeveloperController.php
+++ b/src/Api/Management/Controller/DeveloperController.php
@@ -30,6 +30,7 @@ use Apigee\Edge\Controller\PaginatedEntityIdListingControllerTrait;
 use Apigee\Edge\Controller\PaginatedEntityListingControllerTrait;
 use Apigee\Edge\Controller\PaginationHelperTrait;
 use Apigee\Edge\Controller\StatusAwareEntityControllerTrait;
+use Apigee\Edge\Entity\EntityInterface;
 use Apigee\Edge\Serializer\EntitySerializerInterface;
 use Apigee\Edge\Structure\PagerInterface;
 use Psr\Http\Message\UriInterface;
@@ -64,7 +65,7 @@ class DeveloperController extends PaginatedEntityController implements Developer
     }
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
      */
     public function getDeveloperByApp(string $appName): DeveloperInterface
     {
@@ -83,7 +84,7 @@ class DeveloperController extends PaginatedEntityController implements Developer
     }
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
      */
     public function getEntities(
         PagerInterface $pager = null,
@@ -95,7 +96,24 @@ class DeveloperController extends PaginatedEntityController implements Developer
     }
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
+     *
+     * We are enforcing email addresses over developer ids (UUIDs) when we are updating user profile.
+     *
+     *  @see https://github.com/apigee/apigee-edge-drupal/issues/594
+     */
+    public function update(EntityInterface $entity): void
+    {
+        $uri = $this->getEntityEndpointUri($entity->originalEmail());
+        $response = $this->getClient()->put(
+            $uri,
+            $this->getEntitySerializer()->serialize($entity, 'json')
+        );
+        $this->getEntitySerializer()->setPropertiesFromResponse($response, $entity);
+    }
+
+    /**
+     * {@inheritdoc}
      */
     protected function getBaseEndpointUri(): UriInterface
     {
@@ -104,7 +122,7 @@ class DeveloperController extends PaginatedEntityController implements Developer
     }
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
      */
     protected function getEntityClass(): string
     {
--- a/src/Api/Management/Controller/DeveloperController.php
+++ b/src/Api/Management/Controller/DeveloperController.php
@@ -98,7 +98,7 @@ class DeveloperController extends PaginatedEntityController implements Developer
     /**
      * {@inheritdoc}
      *
-     * We are enforcing email addresses over developer ids (UUIDs) when we are updating user profile.
+     * We are enforcing email addresses over developer-ids (UUIDs) when we are updating user profile.
      *
      * @psalm-suppress PossiblyNullArgument $entity->originalEmail() is not null here.
      *
--- a/src/Api/Management/Controller/DeveloperController.php
+++ b/src/Api/Management/Controller/DeveloperController.php
@@ -100,6 +100,8 @@ class DeveloperController extends PaginatedEntityController implements Developer
      *
      * We are enforcing email addresses over developer ids (UUIDs) when we are updating user profile.
      *
+     * @psalm-suppress UndefinedMethod
+     *
      *  @see https://github.com/apigee/apigee-client-php/issues/153
      */
     public function update(EntityInterface $entity): void
--- a/src/Api/Management/Controller/DeveloperController.php
+++ b/src/Api/Management/Controller/DeveloperController.php
@@ -98,20 +98,22 @@ class DeveloperController extends PaginatedEntityController implements Developer
     /**
      * {@inheritdoc}
      *
-     * We are enforcing email addresses over developer-ids (UUIDs) when we are updating user profile.
+     * We are enforcing email addresses over developer ids (UUIDs) when we are updating user profile.
      *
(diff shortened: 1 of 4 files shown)


src/Entity/Property/EmailPropertyAwareTrait.php
```diff
@@ -24,18 +24,36 @@ trait EmailPropertyAwareTrait
...
         $this->email = $email;
+        if (null === $this->originalEmail) {
+            $this->originalEmail = $email;
+        }
+    }
+
+    /**
+     * The original email address of the developer used when updating user email address.
+     */
+    public function originalEmail(): ?string
```
giteshk: Can this be a private methods?

src/Entity/Property/EmailPropertyAwareTrait.php
```diff
@@ -24,18 +24,36 @@ trait EmailPropertyAwareTrait
     protected $email;
```
giteshk, line 24: Is this trait used anywhere else?

src/Entity/Property/EmailPropertyAwareTrait.php
shishir-intelli: @giteshk  No, this cant be a private/protected method, this is called by `DeveloperController` class from Management API.
This method is only used when updating the user profile and specially when updating users email address.

src/Entity/Property/EmailPropertyAwareTrait.php
shishir-intelli: Yes, this is generic trait for setting/getting email address.

giteshk: Does it need it's own trail is what I am thinking. OriginalEmailPropertyAwareTrait ?

so if someone wanted to use EmailPropertyAwareTrait (e.g. DeveloperApps) trait in the future they would get this property by default even if they didn't want it.

shishir-intelli: @giteshk I removed it from `EmailPropertyAwareTrait` and implemented in `Developer` entity. [Here](https://github.com/apigee/apigee-client-php/pull/154/commits/7c7a4fe24cd89cdac34836ba97813c171584bb6c)
(It is a method just to save the Original email address so trait not required.)

giteshk: This looks better 👍 

Thank you

giteshk (approved):
LGTM .. Pending travis-ci.com migration. Please verify on your travis-ci instance.

src/Api/Management/Controller/DeveloperController.php
```diff
@@ -95,7 +96,28 @@ public function getEntities(
...
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
+     *
+     * We are enforcing email addresses over developer ids (UUIDs) when we are updating user profile.
+     *
+     * @psalm-suppress PossiblyNullArgument $entity->originalEmail() is not null here.
+     *
+     *  @see https://github.com/apigee/apigee-client-php/issues/153
```
mxr576: Nitpick unnecessary space before `@see`

src/Api/Management/Controller/DeveloperController.php
```diff
@@ -104,7 +126,7 @@ protected function getBaseEndpointUri(): UriInterface
     }
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
```
mxr576, line 129: IIRC it was intentionally `@inheritdoc` instead of the inlined version of it. Besides, it is an unrelated change, so I would propose changing this in a separate CR.

src/Api/Management/Entity/Developer.php
```diff
@@ -44,6 +44,13 @@ class Developer extends AppOwner implements DeveloperInterface
     /** @var string[] */
     protected $companies = [];
 
+    /**
+    * The original email address of the developer.
+    *
+    * @var string|null
+    */
+    protected $originalEmail;
```
mxr576: Since here a public getter for this property, this property should be `private`, because no subclass should modify this directly. The related business logic is managed by the `setEmail()` class and both child classes and external caller should only use that public interface.

src/Api/Management/Entity/Developer.php
```diff
@@ -91,58 +98,79 @@ public function setCompanies(string ...$companies): void
...
+            $this->originalEmail = $email;
+        }
+    }
+
+    /**
+     * The original email address of the developer used when updating user email address.
+     *
+     * @see Apigee\Edge\Api\Management\Controller\DeveloperController
+     */
+    public function originalEmail(): ?string
```
mxr576, line 174: Is there a reason for not having "get" as prefix?

mxr576 (changes requested):

src/Api/Management/Entity/Developer.php
shishir-intelli: Yes, implementing it as `GET` causes "`Invalid JSON payload received . Unknown name \"originalEmail\" at 'developer': Cannot find field.`" error and update fails.

src/Api/Management/Controller/DeveloperController.php
shishir-intelli: @mxr576 Updated.
Thanks for noticing it.

src/Api/Management/Entity/Developer.php
shishir-intelli: @mxr576 Fixed,
You are right, Thanks!

src/Api/Management/Controller/DeveloperController.php
shishir-intelli: https://github.com/apigee/apigee-client-php/issues/157

src/Api/Management/Entity/Developer.php
mxr576: Okay, so this happens because the default serialization solution in the library assumes that every public getter has to be serialized and added to the request payload.
A proper solution would be convincing the serializer not to serialize the `getOriginalEmail()`. IIRC there are examples for entity-type specific serializers in the library.

The other way around is keeping this method, marking them as `@internal` (even if it is public) and explaining in a short comment why it is not "get...".

src/Api/Management/Controller/DeveloperController.php
mxr576: Thanks, so please roll back unrelated changes from this PR if possible.

shishir-intelli: @mxr576 if we roll back this change the `php-cs-fixer` throws `phpdoc_tag_type` error, so I think we should let this change go with this PR and for rest `phpdoc_tag_type` issue we can do it in #157

image

PR #158, opened 2021-06-30
[#157] Fixes phpdoc_tag_type and lambda_not_used_import error

### LINKED ISSUE #157: {@inheritdoc} instead of `@inheritdoc` to avoid phpdoc_tag_type error
IIRC it was intentionally `@inheritdoc` instead of the inlined version of it. Besides, it is an unrelated change, so I would propose changing this in a separate CR.

_Originally posted by @mxr576 in https://github.com/apigee/apigee-client-php/pull/154#discussion_r659284343_

phdhiren:


shishir-intelli (approved):
LGTM

PR #159, opened 2021-06-30
Ignore global averages in optimized metrics calculation

mxr576:
## Problem/Motivation

It seems when Apigee can, it calculates global averages for metrics like `avg(total_response_time)`.  See: https://www.googlecloudcommunity.com/gc/Apigee/Analytics-Server-Issue-avg-total-response-time-is-fetching/td-p/58868

The only problem with this that optimizing these global averages  (making sure that there is a result for every time unit in these metrics) does not make any sense.
When they are returned by the API backend they break current implementation with:

```
Error: Unsupported operand types in Apigee\Edge\Api\Management\Controller\StatsController->fillGapsInMetricsData() (line 307 of /mnt/files/local_mount/build/vendor/apigee/apigee-client-php/src/Api/Management/Controller/StatsController.php) 

Warning: array_combine(): Both parameters should have an equal number of elements in Apigee\Edge\Api\Management\Controller\StatsController->fillGapsInMetricsData() (line 306 of /mnt/files/local_mount/build/vendor/apigee/apigee-client-php/src/Api/Management/Controller/StatsController.php) 
```

caused by 

```json
                    "metric": [
                        {
                            "env": "prod",
                            "name": "avg(total_response_time)",
                            "values": [
                                177.16666666666666,
                                183.55555555555554,
                                176.9,
                                153.63636363636363,
                                154,
                                194.71428571428572,
                                165.625,
                                185.0909090909091,
                                166.41666666666666,
                                208.375,
                                179.4
                            ]
                        },
                        {
                            "env": "prod",
                            "name": "global-avg-total_response_time",
                            "values": [
                                177
                            ]
                        }
                    ]
```

## Needs investigation
Is there a way to convince the Stats API not to generate/return this value? See: https://www.googlecloudcommunity.com/gc/Apigee/Analytics-Server-Issue-avg-total-response-time-is-fetching/td-p/58868

### DIFF
--- a/tests/Api/Management/Controller/StatsControllerTest.php
+++ b/tests/Api/Management/Controller/StatsControllerTest.php
@@ -171,7 +171,7 @@ class StatsControllerTest extends ControllerTestBase
 
     protected function emptyResponseArray(): array
     {
-        return ['Response' => ['TimeUnit' => [], 'stats' => ['data' => [0 => ['values' => []]]]]];
+        return ['Response' => ['TimeUnit' => [], 'stats' => ['data' => [0 => ['name' => 'foo', 'env' => 'bar', 'values' => []]]]]];
     }
 
     /**
--- a/src/Api/Management/Controller/StatsController.php
+++ b/src/Api/Management/Controller/StatsController.php
@@ -302,7 +302,18 @@ class StatsController extends AbstractController implements StatsControllerInter
     private function fillGapsInMetricsData(bool $tsAscending, array $allTimeUnits, array $originalTimeUnits, array &$metricsData): void
     {
         $zeroArray = array_fill_keys($allTimeUnits, 0);
+        $needs_reindex = false;
         foreach ($metricsData as $key => $metric) {
+            // Ignore and remove every global average metric from results.
+            // As these are global averages, it does not make sense filling
+            // gaps on those. If this information is needed then the
+            // non-optimized method should be used.
+            // @see https://www.googlecloudcommunity.com/gc/Apigee/Analytics-Server-Issue-avg-total-response-time-is-fetching/td-p/58868
+            if (0 === strpos($metric['name'], 'global-')) {
+                unset($metricsData[$key]);
+                $needs_reindex = true;
+                continue;
+            }
             $metricsData[$key]['values'] = array_combine($originalTimeUnits, $metric['values']);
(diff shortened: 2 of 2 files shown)


giteshk:
@mxr576 Can you add details on this issue.
Ideally we do not want to manipulate the response from the backend API.

mxr576:
@giteshk I have shared everything that I could about this topic, what kind of other input do you think would be useful?

As I can see, this is a regression bug and it is odd that I was the first one who bumped into this problem by using the PHP API client. As you can see from the referenced community post, others also did not under why this `global-` average was returned by the Apigee API backend when it was not requested.

> Ideally we do not want to manipulate the response from the backend API.

But this is what we already do by creating these "optimized" metrics data instead of returning the raw data from the Apigee API backend - that does not contain data for every time unit in the given timeframe, only for those when there was actual "activity". (So basically 0-s are usually not returned by Apigee.)

mxr576:
@giteshk and others, https://github.com/apigee/apigee-edge-drupal/issues/612 is a bug report about the same problem.

maxim8898:
Hi @mxr576 , I applied your patch on my project, the problem is gone!

giteshk:
@raakesh-blokhra Can we review this please?

raakesh-blokhra:
@mxr576 Once the test results are cleared we'll review and approve accordingly.

tests/Api/Management/Controller/StatsControllerTest.php
```diff
@@ -171,7 +171,7 @@ public function testGapFilling(): void
 
     protected function emptyResponseArray(): array
     {
-        return ['Response' => ['TimeUnit' => [], 'stats' => ['data' => [0 => ['values' => []]]]]];
+        return ['Response' => ['TimeUnit' => [], 'stats' => ['data' => [0 => ['name' => 'foo', 'env' => 'bar', 'values' => []]]]]];
```
mxr576, line 174: Fixed mock data based on real-life API responses.

mxr576:
@phdhiren Fixed CS issues and failing tests on the PR, please review.

phdhiren (approved):
Change looks good to me, Thank you @mxr576 for the contribution.

mxr576:
thanks, please merge it when you have some time

PR #164, opened 2021-07-07
[#112] Change ClientInterface constant names and HybridOauth2 class to fit new Apigee naming

### LINKED ISSUE #112: Change ClientInterface constant names and HybridOauth2 class to fit new Apigee naming
The Apigee product used to be hosted at apigee.com. The newer platform of Apigee is now at apigee.google.com and uses Google Cloud. The actual names of these platforms have been changed to:

- apigee.com: Apigee Edge
- apigee.google.com: Apigee on Google Cloud

Hybrid is the name for a specific flavor of Apigee on Google Cloud where the Apigee runtime is hosted in a different location than Google Cloud. However, the Apigee API and admin features are still at apigee.google.com.

Since we did not understand the naming when we created the feature, the constant `ClientInterface::HYBRID_ENDPOINT` is not correct.  In the ClientInteface, the constants should be renamed:

- `ClientInterface::DEFAULT_ENDPOINT` should be `EDGE_ENDPOINT`
- `ClientInterface::HYBRID_ENDPOINT` should be `APIGEE_ON_GCP_ENDPOINT`

To avoid breaking changes, the original constants should be marked as deprecated, and the new constants added until we can remove them in a 3.x version.  Add comments to explain this info above.

The class `HybridOauth2` class is also named incorrectly, and should be deprecated in favor of a class named `ApigeeOnGcpOauth2`

phdhiren:


raakesh-blokhra (approved):
LGTM

PR #169, opened 2021-07-22
ApigeeX code for supported currency

divya-intelli:
ApigeeX code for the supported currency of the organization, currently USD is added as default as ApigeeX does not have supported currency api.

### DIFF
--- a/src/Api/ApigeeX/Controller/ListingHelperTrait.php
+++ b/src/Api/ApigeeX/Controller/ListingHelperTrait.php
@@ -72,20 +72,6 @@ trait ListingHelperTrait
         return $responseArray;
     }
 
-    /**
-     * Returns an array of supported currency.
-     *
-     * @param array $currencyArray
-     *   Supported currency array
-     *
-     * @return array
-     *   Values as an array.
-     */
-    protected function getdefaultCurrency($currencyArray): array
-    {
-        return $this->responseArrayToArrayOfEntities($currencyArray);
-    }
-
     abstract protected function responseToArray(ResponseInterface $response, bool $expandCompatibility = false): array;
 
     abstract protected function responseArrayToArrayOfEntities(array $responseArray, string $keyGetter = 'id'): array;
--- a/src/Api/ApigeeX/Controller/ListingHelperTrait.php
+++ b/src/Api/ApigeeX/Controller/ListingHelperTrait.php
@@ -43,6 +43,7 @@ trait ListingHelperTrait
     protected function getRawList(UriInterface $uri): array
     {
         $response = $this->getClient()->get($uri);
+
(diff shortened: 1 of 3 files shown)


src/Api/ApigeeX/Controller/SupportedCurrencyController.php
```diff
@@ -0,0 +1,83 @@
...
+
+        return $this->getdefaultCurrency($currencyarray);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getBaseEndpointUri(): UriInterface
+    {
+        return 'null';
```
mxr576: How/why this is valid?

mxr576: It feels to me that this change abuses the original architecture, if the Apigee X version of this API does not exist, yet we need something that returns this static value then the original architecture/code needs refactoring, or this implementation should not extend/implement existing classes/interfaces.

src/Api/ApigeeX/Controller/SupportedCurrencyController.php
```diff
@@ -0,0 +1,83 @@
...
+    use PaginatedListingHelperTrait;
+
+    /**
+     * SupportedCurrencyController constructor .
+     *
+     * @param string $organization
+     * @param \Apigee\Edge\ClientInterface $client
+     * @param \Apigee\Edge\Serializer\EntitySerializerInterface|null $entitySerializer
+     */
+    public function __construct(string $organization, ClientInterface $client, ?EntitySerializerInterface $entitySerializer = null)
```
mxr576: IMO, this code possibly incorrectly uses inheritance and possibly violates the interface segregation principle.
This solution which returns a static "supported currency" does not need an API client, neither an (entity)  serializer, it can already create the entity from the static data.

mxr576 (changes requested):

src/Api/ApigeeX/Controller/ListingHelperTrait.php
```diff
@@ -54,6 +55,37 @@ protected function getRawList(UriInterface $uri): array
...
+    /**
+     * Returns an array of supported currency.
+     *
+     * @param array $currencyArray
+     *   Supported currency array
+     *
+     * @return array
+     *   Values as an array.
+     */
+    protected function getdefaultCurrency($currencyArray): array
```
mxr576: Nitpick: `getDefaultCurrency`, but this trait is used by other entity classes (like `\Apigee\Edge\Api\ApigeeX\Controller\AcceptedRatePlanController`) so this method which is only used by SuppertedCurrency controller must not be here.

raakesh-blokhra (approved):
LGTM

PR #171, opened 2021-07-30
Updated user-agent string with additional information

### LINKED ISSUE #556: Add Drupal core version to the user agent string
## Description
When calls are made to Apigee APIs, the user agent string sends info about the Apigee Edge version and PHP Client version:

`Apigee Edge DevPortal 8.x-1.10 (Apigee Edge PHP Client 2.0.3)`

This task is to add the Drupal core version along with this information:

`Drupal Core 8.19, Apigee Edge DevPortal 8.x-1.10 (Apigee Edge PHP Client 2.0.3)`

shishir-intelli:
Closes https://github.com/apigee/apigee-edge-drupal/issues/556

To implement the UA in the below format :
`Apigee Edge/1.22 (Apigee Edge PHP Client/2.0.3; Drupal/8.9.16; PHP/7.3.11)`
and
`Apigee Monetization/1.10 (Apigee Edge/8.x-1.22; Apigee Edge PHP Client/2.0.3; Drupal/8.9.16; PHP/7.3.11)`

We had to change the `getUserAgent()`, here we are providing more information about Drupal core versions and modules version by exploding the string passed as parameter by Apigee_edge module `userAgentPrefix()`.

Need to make sure about the changes made in library in getUserAgent() function.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -172,9 +172,15 @@ class Client implements ClientInterface
     public function getUserAgent(): string
     {
         if (null !== $this->userAgentPrefix) {
+            $userAgent = explode('~', $this->userAgentPrefix);
+            if (count($userAgent) > 1) {
+                if ("" === $userAgent[1]) {
+                    return sprintf("{$userAgent[0]} ({$this->getClientVersion()}; {$userAgent[2]}; {$this->getPHPVersion()})");
+                }
+                return sprintf("{$userAgent[1]} ({$userAgent[0]}; {$this->getClientVersion()}; {$userAgent[2]}; {$this->getPHPVersion()})");
+            }
             return sprintf("{$this->userAgentPrefix} ({$this->getClientVersion()})");
         }
-
         return $this->getClientVersion();
     }
 
@@ -183,7 +189,15 @@ class Client implements ClientInterface
      */
     public function getClientVersion(): string
     {
-        return sprintf('Apigee Edge PHP Client %s', self::VERSION);
+        return sprintf('Apigee Edge PHP Client/%s', self::VERSION);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPHPVersion(): string
+    {
+        return sprintf('PHP/%s', phpversion());
     }
 
     /**
--- a/src/Client.php
+++ b/src/Client.php
@@ -171,11 +171,12 @@ class Client implements ClientInterface
      */
     public function getUserAgent(): string
     {
+        $user_agent = $this->getClientVersion() . '; PHP/' . PHP_VERSION;
         if (null !== $this->userAgentPrefix) {
-            return $this->userAgentPrefix . ' (' . $this->getClientVersion() . '; PHP/' . PHP_VERSION . ')';
+            return $this->userAgentPrefix . ' (' . $user_agent . ')';
         }
 
-        return $this->getClientVersion() . '; PHP/' . PHP_VERSION;
+        return $user_agent;
     }
 
     /**
--- a/src/Client.php
+++ b/src/Client.php
@@ -172,20 +172,10 @@ class Client implements ClientInterface
     public function getUserAgent(): string
     {
         if (null !== $this->userAgentPrefix) {
-            // Checks if userAgent has multiple values in string.
-            $userAgent = explode('~', $this->userAgentPrefix);
-            if (count($userAgent) > 1) {
-                if ('' === $userAgent[1]) {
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -183,7 +192,15 @@ public function getUserAgent(): string
...
     public function getClientVersion(): string
     {
-        return sprintf('Apigee Edge PHP Client %s', self::VERSION);
+        return sprintf('Apigee Edge PHP Client/%s', self::VERSION);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPHPVersion(): string
```
mxr576: Why this has to be public? Why a new method is needed for this simple system call?

src/Client.php
```diff
@@ -172,6 +172,15 @@ public function getEndpoint(): string
     public function getUserAgent(): string
     {
         if (null !== $this->userAgentPrefix) {
+            $userAgent = explode('~', $this->userAgentPrefix);
+            if (count($userAgent) > 1) {
+                if ('' === $userAgent[1]) {
+                    return sprintf("{$userAgent[0]} ({$this->getClientVersion()}; {$userAgent[2]}; {$this->getPHPVersion()})");
+                }
+
+                return sprintf("{$userAgent[1]} ({$userAgent[0]}; {$this->getClientVersion()}; {$userAgent[2]}; {$this->getPHPVersion()})");
```
mxr576: Incorrect usage of `sprintf()`, `{$foo}` works without `sprintf()`, Use variable replacements, then `$this->getPHPVersion()` method also becomes unnecesary.)

mxr576 (changes requested):

mxr576:
"Add Drupal core version" is unrelated to the topic of this repository in PR's title, I would remove that. Also, commit msg/PR title does not have to end with a full stop.

phdhiren (changes requested):
Instead of said required format it can be little more simple - Inside bracket client library related information i.e. apigee-client-php and PHP version. Other information should be prefixed. This should help other consumer of apigee-client-php efficiently and not require to specific version match.

src/Client.php
```diff
@@ -172,16 +172,17 @@ public function getEndpoint(): string
...
+            // Checks if userAgent has multiple values in string.
             $userAgent = explode('~', $this->userAgentPrefix);
             if (count($userAgent) > 1) {
                 if ('' === $userAgent[1]) {
-                    return sprintf("{$userAgent[0]} ({$this->getClientVersion()}; {$userAgent[2]}; {$this->getPHPVersion()})");
+                    return "{$userAgent[0]} ({$this->getClientVersion()}; {$userAgent[2]}; PHP/" . phpversion() . ")";
                 }
 
-                return sprintf("{$userAgent[1]} ({$userAgent[0]}; {$this->getClientVersion()}; {$userAgent[2]}; {$this->getPHPVersion()})");
+                return "{$userAgent[1]} ({$userAgent[0]}; {$this->getClientVersion()}; {$userAgent[2]}; PHP/" . phpversion() . ")";
```
mxr576: Use `PHP_VERSION` instead.

https://github.com/search?l=PHP&q=PHP_VERSION&type=Code
https://github.com/search?q=%22phpversion%28%29%22+language%3APHP&type=Code&ref=advsearch&l=PHP&l=

https://www.php.net/manual/en/reserved.constants.php#reserved.constants.core

mxr576 (changes requested):

src/Client.php
shishir-intelli: Thanks @mxr576 ,
Yes, missed `PHP_VERSION`.
Also removed the complex user-agent string code and instead will send it from dependent modules
according to this comment https://github.com/apigee/apigee-edge-drupal/issues/556#issuecomment-894076924

src/Client.php
```diff
@@ -172,20 +172,10 @@ public function getEndpoint(): string
...
-
-                return "{$userAgent[1]} ({$userAgent[0]}; {$this->getClientVersion()}; {$userAgent[2]}; PHP/" . phpversion() . ")";
-            }
-
-            return "{$this->userAgentPrefix} ({$this->getClientVersion()})";
+            return $this->userAgentPrefix . ' (' . $this->getClientVersion() . '; PHP/' . PHP_VERSION . ')';
         }
 
-        return $this->getClientVersion();
+        return $this->getClientVersion() . '; PHP/' . PHP_VERSION;
```
mxr576: One small nitpick, `$this->getClientVersion() . '; PHP/' . PHP_VERSION` is duplicated, it can be extracted to a new `$user_agent` variable in this method and used in all occurrences of this string.

mxr576 (approved):

phdhiren (approved):
LGTM

PR #175, opened 2021-08-31
Removed unnecessary code which was causing issue while testing for Apigee X

shishir-intelli:
Related to #https://github.com/apigee/apigee-m10n-drupal/issues/306

Removed unnecessary code which were causing issues for running test-case for Apigee X.


raakesh-blokhra (approved):
LGTM

PR #212, opened 2022-03-14
GitHub Actions for Apigee PHP Client

giteshk:
Migrate from circle ci -> github actions
Upload code coverage report
Removed circle ci integration

### DIFF
--- a/.github/workflows/php-test-workflow.yml
+++ b/.github/workflows/php-test-workflow.yml
@@ -19,10 +19,10 @@ jobs:
                     dependencies-preference: " "
                 -   php: '8.0'
                     dependencies-preference: "--prefer-lowest"
-#                -   php: '8.1'
-#                    dependencies-preference: " "
-#                -   php: '8.1'
-#                    dependencies-preference: "--prefer-lowest"
+                -   php: '8.1'
+                    dependencies-preference: " "
+                -   php: '8.1'
+                    dependencies-preference: "--prefer-lowest"
             fail-fast: false
         steps:
         -   uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


giteshk:
GitHub actions tests => https://github.com/giteshk-org/apigee-client-php/actions/runs/1981559088

.github/workflows/php-test-workflow.yml
```diff
@@ -0,0 +1,78 @@
...
+            fail-fast: false
+        steps:
+        -   uses: actions/checkout@v2
+            with:
+                fetch-depth: 0
+        -   name: Setup PHP
+            uses: shivammathur/setup-php@v2
+            with:
+                coverage: "xdebug"
+                ini-values: date.timezone=Europe/Paris,memory_limit=-1,default_socket_timeout=10,session.gc_probability=0,apc.enable_cli=1,zend.assertions=1
```
phdhiren: What is the reason to add date.timezone and why to lower down default_socket_timeout?

.github/workflows/php-test-workflow.yml
```diff
@@ -0,0 +1,78 @@
...
+    pull_request:
+
+jobs:
+    build:
+        name: Tests
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                include:
+                -   php: '7.4'
```
phdhiren, line 14: Are we planning to raise minimum php version to 7.4 instead of 7.3 (which is already EOL)? In that case we should change in composer.json

phdhiren (approved):
LGTM

.github/workflows/php-test-workflow.yml
giteshk: composer.json has been updated

PR #230, opened 2022-07-22
Bump firebase/php-jwt version

### LINKED ISSUE #229: Update firebase/php-jwt depedency
Hello

6.x version of firebase/php-jwt released a few months ago. But this project has support only for 5.x. So we are facing compatibility issues with packages which require the 6.x
Can we make this project  compatible with 6.x version of firebase/php-jwt.

Thanks

divya-intelli:
Closes #229 
Updated firebase/php-jwt version to ^6.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "ext-openssl": "*",
         "ext-reflection": "*",
         "fightbulc/moment": "^1.26",
-        "firebase/php-jwt": "^5.0",
+        "firebase/php-jwt": "^6.0",
         "league/period": "^4.12",
         "php-http/client-common": "^2.0",
         "php-http/client-implementation": "^1.0",


composer.json
```diff
@@ -18,7 +18,7 @@
         "ext-openssl": "*",
         "ext-reflection": "*",
         "fightbulc/moment": "^1.26",
-        "firebase/php-jwt": "^5.0",
+        "firebase/php-jwt": "^6.0",
```
phdhiren, line 21: How are we planning to take care of breaking changes mentioned at https://github.com/firebase/php-jwt/releases/tag/v6.0.0

divya-intelli: From the breaking changes list, we are only using JWT::encode and the syntax used is as mentioned in the breaking changes list i.e. we are passing the third argument.

phdhiren (approved):
LGTM

PR #248, opened 2023-03-03
Support assigning API products to Apps on creation

### LINKED ISSUE #214: Support assinging API products to Apps on creation
**Is your feature request related to a problem? Please describe.**
The API has support for this:
https://apidocs.apigee.com/docs/company-apps/1/routes/organizations/%7Borg_name%7D/companies/%7Bcompany_name%7D/apps/post
https://apidocs.apigee.com/docs/developer-apps/1/routes/organizations/%7Borg_name%7D/developers/%7Bdeveloper_email%7D/apps/post

**Describe the solution you'd like**
The current implementation has on object to cover read/write models for apps and it mostly consider persisted apps. Now we might need to introduce new models for new (non-persisted) apps.

(I remember that there was some design decision that lead to this design but not supporting this might have been an oversight.)

**Describe alternatives you've considered**
-

**Additional context**
-

phdhiren:
Fixes #214


src/Api/Management/Entity/AppInterface.php
```diff
@@ -83,4 +83,12 @@ public function setCallbackUrl(string $callbackUrl): void;
...
      */
     public function getCredentials(): array;
+
+    /**
+     * Set Initial API Product while creating App.
+     *
+     * @param array $initialApiProducts
+     *
+     */
+    public function setInitialApiProducts(array $initialApiProducts): void;
```
mxr576: Generic advice: changing interfaces MUST NOT happen in minor/patch releases. 

Keeping this here with a `@todo` _could_ be a workaround...

src/Api/Management/Entity/App.php
```diff
@@ -216,4 +219,20 @@ public function setScopes(string ...$scopes): void
...
+     */
+    public function getApiProducts(): array
+    {
+        return $this->initialApiProducts;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setInitialApiProducts(array $initialApiProducts): void
```
mxr576: What should/would happen if this is called on an existing entity before a PUT?

src/Api/Management/Entity/App.php
```diff
@@ -216,4 +219,20 @@ public function setScopes(string ...$scopes): void
     {
         $this->privateSetScopes(...$scopes);
     }
+
+    /**
+     * @param array $apiProducts
+     */
+    public function getApiProducts(): array
+    {
+        return $this->initialApiProducts;
```
mxr576, line 232: Returning an empty array here would not be misleading? Since API Product association is stored on credentials and not on the app.

src/Api/Management/Entity/App.php
phdhiren: This should be called only on the creation of the App. 
https://github.com/apigee/apigee-edge-drupal/blob/3e5dcb11906aa49b3499098208c564ab5dedfcd1/src/Entity/Form/AppForm.php#L129

src/Api/Management/Entity/App.php
phdhiren: Agreed. With `getApiProducts` it will post to Edge and `apiProducts` is an IN parameter but in the return we will get inside the credential so there is no `setApiProducts`. Only for App creation will have 'apiProducts', not for the update

src/Api/Management/Entity/App.php
mxr576: Should be, but technically speaking nothing prevents it to be called on an existing app by a developer. Unless you throw an exception here... but this desing goes already againts interface the interface separation principle.

Being that said, I have no better idea than throwing an exception. Everything else would come with bigger refactoring - which I am unsure how long we can avoid...

giteshk (approved):

mxr576:
Sorry, but I still have concerns with this design change that are not answered yet
https://github.com/apigee/apigee-client-php/pull/248/files#r1129955554

src/Api/Management/Entity/App.php
giteshk: @phdhiren is going to address this. Thank you for providing feedback.

phdhiren:
@giteshk @mxr576 

I've added the exception so the function gets called only to create an app and throw for the update app.

src/Api/Management/Entity/App.php
```diff
@@ -216,4 +219,33 @@ public function setScopes(string ...$scopes): void
...
+     * @param array $initialApiProducts
+     *
+     * @throws \Exception If used to update existing App.
+     */
+    final public function setInitialApiProducts(array $initialApiProducts): void
+    {
+        if (!$this->appId) {
+            $this->initialApiProducts = $initialApiProducts;
+        } else {
+            throw new \Exception('This method is only supported for creating a new app.');
```
mxr576: Should be a \LogicException() since calling it is a "mistake made by the developer". 

(Although we know this method should only exist on New app entities for which we do not have an object type...)

mxr576:
Thanks. What about adding test coverage?

tests/Api/Management/Controller/AppControllerTestBase.php
```diff
@@ -116,6 +117,8 @@ protected function alterObjectsBeforeCompareResponseAndCreatedEntity(\stdClass &
     {
         /* @var \Apigee\Edge\Api\Management\Entity\DeveloperAppInterface $created */
         $responseObject->appFamily = $created->getAppFamily();
+
+        $responseObject->apiProducts = $created->getApiProducts();
```
mxr576, line 121: I am not sure if this change covers the fixed issues with a test.

src/Api/Management/Entity/App.php
phdhiren: This has been taken care. Thank you for pointing it out.

tests/Api/Management/Controller/AppControllerTestBase.php
phdhiren: Added separate ticket to test coverage

Issue #258, opened 2023-04-27
Create a 2.1.x branch for Symfony 6.2 compatiblity fixes


giteshk:
Drupal 10 uses Symfony 6.2 , which is not compatible with Drupal 9. 

We will release the Symfony 6.2 changes on Apigee Client PHP in 2.1.0 release.

After 2.1.0, minimal updates will be released in the 2.0.x releases, this is to give customers some time to plan upgrades to Drupal 10.

mxr576:
2.1.x is not going to be good enough because even Apigee Edge Drupal modules [uses](https://github.com/apigee/apigee-edge-drupal/blob/2.0.11/composer.json#LL9C1-L9C1) ^2.0.19 which means >=2.0.0 && <3.0.0 - so Composer is going to install versions with potential BC breaks, see #269. 

My suggestion is opening the 3.x branch for this.

divya-intelli:
@giteshk need your suggestion here.
@mxr576 our current focus is  Drupal 10 compatibility of Apigee modules. As suggested, we will be testing for BC break.

mxr576:
> our current focus is Drupal 10 compatibility of Apigee modules

Ours too, for sure! :) However, we still have to warrant that we are not going to break existing projects. The PHP API client is not only used by the Drupal Apigee Edge module.

By using https://github.com/Roave/BackwardCompatibilityCheck  I have got the following report for comparison between the latest tagged release in 2.x and the code in drupal10-prep branch. Not looking good :( 

```
[BC] CHANGED: The return type of Apigee\Edge\Api\Monetization\NameConverter\NameConverterBase#normalize() changed from no type to string
[BC] CHANGED: The return type of Apigee\Edge\Api\Monetization\NameConverter\NameConverterBase#denormalize() changed from no type to string
[BC] CHANGED: The return type of Apigee\Edge\Api\Monetization\NameConverter\NameConverterBase#normalize() changed from no type to string
[BC] CHANGED: The return type of Apigee\Edge\Api\Monetization\NameConverter\NameConverterBase#denormalize() changed from no type to string
[BC] CHANGED: The return type of Apigee\Edge\Api\Monetization\NameConverter\NameConverterBase#normalize() changed from no type to string
[BC] CHANGED: The return type of Apigee\Edge\Api\Monetization\NameConverter\NameConverterBase#denormalize() changed from no type to string
[...]

PR #270, opened 2023-06-12
Rolled back the changes because of backwards-incompatible

shishir-intelli:
#265
Issue ref : https://github.com/apigee/apigee-client-php/issues/258#issuecomment-1571881053

Deprecated class `Http\Client\HttpClient` will be replaced with `Psr\Http\Client\ClientInterface` in 3.x branch


giteshk (approved):
LGTM

PR #294, opened 2023-08-18
Extended AppGroup class from base class AppOwner

### LINKED ISSUE #293: Extend AppGroup entity from AppOwner class
## Description
Extend AppGroup entity from AppOwner class which is a base class for appgroup entities instead of Entity class.

## Version Info
2.x and 3.x

shishir-intelli:
Closes #293


divya-intelli (approved):
LGTM

PR #310, opened 2023-09-20
Mark psr/http-message >=2.0 supported

mxr576:
Motivation: Make installation on Apigee Edge module easier on Drupal 10+ projects was built with https://packagist.org/packages/drupal/core-recommended#10.1.3 that depends on https://packagist.org/packages/symfony/psr-http-message-bridge#v2.3.1.

What is the problem? 

When a new clean Drupal 10+ project is created with `drupal/core-recommended` Composer picks the highest version of packages, including psr/http-message and it installs 2.0.0. Later when drupal/apigee_edge module gets installed, Composer needs convincing to downgrade to a previous version of psr/http-message because this package needs that. (Luckily the rest of the dependency chain also support 1.x.)

Related:
* https://www.php-fig.org/psr/psr-7/meta/#72-type-additions --> I was uncertain if we would like to push for parameter typehints or not and symfony/psr-http-message-bridge support 1.0, so I kept that lowest boundery.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "php-http/message": "^1.13",
         "php-http/message-factory": "^1.0",
         "phpdocumentor/reflection-docblock": "^5.0",
-        "psr/http-message": "^1.0",
+        "psr/http-message": "^1.0 || ^2.0",
         "symfony/options-resolver": "^6.3",
         "symfony/property-access": "^6.3",
         "symfony/property-info": "^6.3",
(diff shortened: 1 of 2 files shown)


divya-intelli:
Thank you @mxr576 for the PR and the description.

Verified that Drupal 10 uses psr/http-message ^2.0 but due to apigee-client-php still using ^1.0, drupal downgrades psr/http-message to ^1.0 .

For this PR Github action is failing with 2 error https://github.com/apigee/apigee-client-php/actions/runs/6247998648/job/16961690725?pr=310#step:10:666. Once the test are passed we can merge the PR.

mxr576:
I'll get back to this PR, hopefully in the next 48 hours :crossed_fingers:

mxr576:
@divya-intelli all green, back to you! :)

divya-intelli (approved):
LGTM

PR #329, opened 2023-12-05
Fix error when AppGroup list is empty or do not returns appgroups array - 2x

shishir-intelli:
Closes https://github.com/apigee/apigee-client-php/issues/325 for 2x


divya-intelli (approved):
LGTM

PR #334, opened 2023-12-14
Change exceptions used to catch failed GCE authentication

### LINKED ISSUE #333: \Apigee\Edge\HttpClient\Plugin\Authentication\GceServiceAccount::isAvailable() throws exception when called on non-GCE context
## Description

The try/catch inside the mentioned method does not work because a different exception is thrown (on Drupal 10.1) that is not a subclass of the currently catched exception.

## Steps to Reproduce
Build and initiate a GceServiceAccount and try to authenticated with that in a non-GCE hosted environment.

## Actual Behavior
Exception is thrown.

## Expected Behavior
FALSE is returned.

## Screenshots

image

## Version Info
3.0.3

mxr576:
Closes #333

### DIFF
--- a/src/HttpClient/Plugin/Authentication/GceServiceAccount.php
+++ b/src/HttpClient/Plugin/Authentication/GceServiceAccount.php
@@ -21,8 +21,9 @@ namespace Apigee\Edge\HttpClient\Plugin\Authentication;
 use Apigee\Edge\Client;
 use Apigee\Edge\ClientInterface;
 use Apigee\Edge\Exception\ApigeeOnGcpOauth2AuthenticationException;
-use Http\Client\Exception;
 use Http\Message\Authentication\Header;
+use Psr\Http\Client\ClientExceptionInterface;
+use Psr\Http\Client\NetworkExceptionInterface;
 
 /**
  * GCE Service Account authentication plugin for authenticating to Google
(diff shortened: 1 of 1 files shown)


mxr576:
> Alternative considering, but it feels an overkill to me, catch \Psr\Http\Client\ClientExceptionInterface

Maybe this approach combines best from both worlds, isAvailable() should only fail with `NetworkExceptionInterface`, when the project is not hosted on GCE  -- right? However getAccessToken() can also fail when the service account is not configured properly. so the host was reachable and it returned an HTTP >= 400 code.

src/HttpClient/Plugin/Authentication/GceServiceAccount.php
```diff
@@ -61,7 +62,7 @@ public function isAvailable(): bool
             $this->authClient()->get('');
 
             return true;
-        } catch (Exception $e) {
+        } catch (NetworkExceptionInterface $e) {
```
shishir-intelli, line 65: @mxr576 ,
With `NetworkExceptionInterface` I am not able to access the `admin/config/apigee-edge/settings` page for credential settings.

Screenshot :
image

mxr576: Thanks for checking, I think I have found the root cause of that: https://github.com/apigee/apigee-edge-drupal/pull/1019

shishir-intelli: @mxr576 Thankyou

shishir-intelli (approved):
LGTM

PR #365, opened 2024-05-22
Restricted Symfony update to 6.4.6 due to issue in latest version 6.4.7

### LINKED ISSUE #361: Update symfony minimum version to 6.4
## Description
As the Symfony 6.3 version support has reached its end of support and a new symfony version 6.4 is released which will be a long-term support till Nov. 2027, we will update the minimum requirement to ^6.4.*

Original Issue - https://github.com/symfony/symfony/issues/54983 - Symfony Serializer adding a conflict as 6.4.7creates above mentioned issue

kedarkhaire:
Fixes #361

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,10 +28,17 @@
         "php-http/message-factory": "^1.0",
         "phpdocumentor/reflection-docblock": "^5.0",
         "psr/http-message": "^1.0 || ^2.0",
-        "symfony/options-resolver": "^6.3",
-        "symfony/property-access": "^6.3",
-        "symfony/property-info": "^6.3",
-        "symfony/serializer": "^6.3.11"
+        "symfony/options-resolver": "6.4.*",
+        "symfony/property-access": "6.4.6",
+        "symfony/property-info": "6.4.*",
+        "symfony/string": "6.4.*",
+        "symfony/var-exporter": "6.4.6",
+        "symfony/stopwatch":  "6.4.*",
+        "symfony/finder": "6.4.*",
+        "symfony/filesystem": "6.4.6",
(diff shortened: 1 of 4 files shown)


divya-intelli (commented):
Please don't restrict the version.

composer.json
```diff
@@ -28,10 +28,17 @@
...
-        "symfony/options-resolver": "^6.3",
-        "symfony/property-access": "^6.3",
-        "symfony/property-info": "^6.3",
-        "symfony/serializer": "^6.3.11"
+        "symfony/options-resolver": "6.4.*",
+        "symfony/property-access": "6.4.6",
+        "symfony/property-info": "6.4.*",
+        "symfony/string": "6.4.*",
+        "symfony/var-exporter": "6.4.6",
+        "symfony/stopwatch":  "6.4.*",
```
divya-intelli: Do we need this?

mxr576:
If there is an issue with symfony/[package] 6.4.7, just add a `conflict` rule to package(s) and that version, nothing more.

mxr576:
Also referencing the related issue - at least here -from Symfony's issue queue would be beneficial.

composer.json
kedarkhaire: Removed

mxr576:
> Also referencing the related issue - at least here -from Symfony's issue queue would be beneficial.

Could this be related to this?

https://github.com/symfony/symfony/issues/54983

kedarkhaire:
Yes @mxr576 that can be the related to the mentioned issue as only `symfony/serializer` with version 6.4.7 is causing issue.

divya-intelli (approved):
LGTM

PR #376, opened 2024-09-11
Remove symfony/serializer conflict and bump version

### LINKED ISSUE #373: Unable to install 3.0.6 on Drupal 10.3.2 or on any version that requires symfony/serializer:>6.4.6
## Description
There was an issue with Symfony Serializer 6.4.7 but that was a long time ago and since https://github.com/apigee/apigee-client-php/commit/3073f76dff563f7de88487697a6b218b7d6adc1f was merged Symfony released several patch versions for 6.4.x branch and 6.4.9 has the issue fixed: https://github.com/symfony/symfony/pull/57580

Because `drupal/core-recommended` locks on specific version of dependencies, the Apigee Edge Drupal module cannot be installed on Drupal 10.3.2 (or on any other version) that requires a never version of Symfony Serializer.

Next time before a release, it should checked if there is a never version from the dependency that was locked to a specific version range or a conflict rule was added and whether that version addresses the reason behind the version lock or not.

## Steps to Reproduce

`composer create-project drupal/recommended-project:~10.3.0`

```sh
$ composer req drupal/apigee_edge:~3.0.8
./composer.json has been updated

  Problem 1
    - Root composer.json requires drupal/apigee_edge ~3.0.8 -> satisfiable by drupal/apigee_edge[3.0.8].
    - apigee/apigee-client-php 3.0.6 conflicts with symfony/serializer >6.4.6.
    - drupal/core-recommended 10.3.2 requires symfony/serializer ~v6.4.7 -> satisfiable by symfony/serializer[v6.4.10].
    - drupal/apigee_edge 3.0.8 requires apigee/apigee-client-php ~3.0.6 -> satisfiable by apigee/apigee-client-php[3.0.6].
    - drupal/core-recommended is locked to version 10.3.2 and an update of this package was not requested.

```

## Actual Behavior
What happened after the steps to reproduce.

## Version Info
In what version or versions of this library can you
reproduce this issue?

shishir-intelli:
Closes #373

In `symfony/serializer:>6.4.6` 

> $context['_read_attributes'] = false;

 prevents the attributes' values from being read if the attribute setters are not implemented.
Ref : https://github.com/symfony/symfony/blob/5df912c9786c53f413c2604b3f979461a36486c5/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php#L312

Now, we added setters for attributes that were previously not included because of this
https://github.com/symfony/property-access/blob/v4.2.5/PropertyAccessor.php#L433-L435

### DIFF
--- a/src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
+++ b/src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
@@ -22,7 +22,7 @@ use Apigee\Edge\Structure\ObjectCopyHelperTrait;
 
 /**
  * Base class for all supported Monetization reports.
- * 
+ *
  * @internal
  *
  * @see https://docs.apigee.com/api-platform/monetization/create-reports#repdefconfigapi
@@ -119,7 +119,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$appIds
      *
-     * @return \self
+     * @return self
      */
     public function setApps(string ...$appIds): self
     {
@@ -139,7 +139,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$currencyIds
      *
-     * @return \self
+     * @return self
      */
     public function setCurrencies(string ...$currencyIds): self
     {
@@ -159,7 +159,7 @@ abstract class AbstractCriteria
     /**
      * @param string|null $currencyOption
      *
-     * @return \self
+     * @return self
      */
     public function setCurrencyOption(?string $currencyOption): self
     {
@@ -173,7 +173,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$developerIds
      *
-     * @return \self
+     * @return self
      */
     public function setDevelopers(string ...$developerIds): self
     {
@@ -201,7 +201,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$apiPackageIds
      *
-     * @return \self
+     * @return self
      */
     public function setApiPackages(string ...$apiPackageIds): self
     {
@@ -221,7 +221,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$apiProductIds
      *
-     * @return \self
+     * @return self
      */
     public function setApiProducts(string ...$apiProductIds): self
     {
@@ -241,7 +241,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$pricingTypes
      *
-     * @return \self
+     * @return self
      */
     public function setPricingTypes(string ...$pricingTypes): self
     {
@@ -261,7 +261,7 @@ abstract class AbstractCriteria
     /**
      * @param string ...$ratePlanLevels
      *
-     * @return \self
+     * @return self
      */
     public function setRatePlanLevels(string ...$ratePlanLevels): self
     {
@@ -305,7 +305,7 @@ abstract class AbstractCriteria
     /**
      * @param bool $show
      *
-     * @return \self
+     * @return self
      */
     public function setShowRevenueSharePercentage(bool $show): self
     {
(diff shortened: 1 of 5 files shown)


src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
```diff
@@ -125,9 +119,9 @@ public function getApps(): array
     /**
      * @param string ...$appIds
      *
-     * @return \self
+     * @return self
      */
-    public function apps(string ...$appIds): self
+    public function setApps(string ...$appIds): self
```
mxr576, line 139: Since the original method names were public APIs, I would suggest  keeping those and may triggering a deprecation error and remove them in the next major version.
I would not just rename (and with that remove) them.

mxr576 (changes requested):

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
shishir-intelli: @mxr576 
Thank you for the suggestion. 
I agree that keeping the original method names and triggering a deprecation warning is a prudent approach. 
We can then plan to remove the deprecated methods in the next major version. I'll make the necessary changes and update the documentation accordingly.

kedarkhaire (approved):
LGTM

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
```diff
@@ -125,7 +119,10 @@ public function getApps(): array
     /**
      * @param string ...$appIds
      *
-     * @return \self
+     * @return self
+     *
+     * @deprecated in 3.0.7, will be removed in 4.0.0. No longer needed.
+     * https://github.com/apigee/apigee-client-php/issues/373
      */
     public function apps(string ...$appIds): self
```
mxr576, line 127: I'd call the new method in the deprecated method for DRY.

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
```diff
@@ -125,7 +119,10 @@ public function getApps(): array
     /**
      * @param string ...$appIds
      *
-     * @return \self
+     * @return self
+     *
+     * @deprecated in 3.0.7, will be removed in 4.0.0. No longer needed.
```
mxr576, line 124: Don't we also have a habit to  trigger runtime deprecation via trigger_error()?

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
shishir-intelli: like this inside deprecated methods?

> $this->apps = setApps($appIds);

or

> return setApps($appIds);

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
shishir-intelli: Yes we do.
Okay let me update the PR.

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
```diff
@@ -125,9 +119,24 @@ public function getApps(): array
...
+     * https://github.com/apigee/apigee-client-php/issues/373
      */
     public function apps(string ...$appIds): self
+    {
+        trigger_error(__METHOD__ . ' is deprecated in 3.0.7, will be removed in 4.0.0: use setApps() instead.', E_USER_DEPRECATED);
```
mxr576, line 129: I'd use `@trigger_error()`, but maybe that is just my preference.

mxr576 (approved):

src/Api/Monetization/Structure/Reports/Criteria/AbstractCriteria.php
shishir-intelli: Thanks @mxr576 
Noted, We will take care of it next time.

PR #407, opened 2025-01-20
Removed PHP8.1 test from matrix for 4x

kedarkhaire:
Removed test check with PHP 8.1 matrix as symfony 7 minimum requirement is 8.2


divya-intelli (approved):
LGTM

PR #419, opened 2025-08-28
Added DRZ support for Drupal 10

### LINKED ISSUE #405: Support DRZ-Enabled Apigee X Organizations
**Is your feature request related to a problem? Please describe.**

Currently the client only supports the global apigee endpoint `apigee.googleapis.com`

**Describe the solution you'd like**

I'd like to be able to use apigee drz endpoints (e.g. `us-apigee.googleapis.com`) in the portal connection settings.

**Describe alternatives you've considered**

None.

**Additional context**

This is a critical request due to the fact that the Integrated Developer Portal isn't supported for DRZ-enabled Apigee X organizations.

kedarkhaire:
Added DRZ support for Drupal 10 Fixes #405


src/ClientInterface.php
shishir-intelli, line 1: @giteshk 
Are we planning to implement region based endpoint in this way, by adding endpoint for all locations?

shishir-intelli: Is this the correct approach, or should the endpoints be handled dynamically based on locations?

giteshk: This would incorrectly handle private cloud URLs

kedarkhaire: Updated as per the pointers

divya-intelli:
Added in #420 , hence closing.

PR #422, opened 2025-09-11
Added endpoint to update scope when apiproduct is added

divya-intelli:
Fixes [#1174](https://github.com/apigee/apigee-edge-drupal/issues/1174)
Added AppGroup scope override endpoint.


shishir-intelli (approved):
LGTM

Issue #430, opened 2025-09-22
Add 'email' field to AppGroup details on PUT call (AppGroup Mint)


shishir-intelli:
**Is your feature request related to a problem? Please describe.**
We need to update the AppGroups PUT API call. The request body must be modified to include the email field within the `__apigee_reserved__developer_details` object.

**Additional context**
ApigeeX

PR #438, opened 2025-09-29
Added 'email' field to AppGroup details on PUT call 4.x

shishir-intelli:
Closes https://github.com/apigee/apigee-client-php/issues/430
ApigeeX AppGroup Mint
This PR is for 4x branch -Drupal 11


shishir-intelli:
See PR https://github.com/apigee/apigee-client-php/pull/439

PR #454, opened 2025-12-23
support firebase/php-jwt ^7.0 | 3.x

shailesh-google:
security: upgrade firebase/php-jwt to latest v7.0.2

This update resolves the security vulnerability reported in [#1204](https://github.com/apigee/apigee-edge-drupal/issues/1204) regarding weak key validation. The library constraint is now expanded to support from v6 to v7.


divya-intelli (approved):
LGTM

PR #462, opened 2026-02-04
 Resolved Developer test failure and Upgrade Symfony components to 7.4

shailesh-google:
Fixed a regression in `DeveloperControllerTest` caused by the missing 'originalEmail' field in the API response by adjusting the test comparison logic for serializer versions >= 7.3.11. Upgraded property-access, options-resolver, and property-info to 7.4. 

Fix: PR #466


shishir-intelli (approved):
LGTM

PR #463, opened 2026-02-05
ignore originalEmail attribute and fix test cases | 3.x

shailesh-google:
The 'originalEmail' attribute was causing a bad attribute error in Apigee X api during user creation from drupal side. This commit adds 'originalEmail' to the ignored attributes list during Developer entity serialization.

It also updates the test helper methods (alterArrays and alterObjects) to account for changes in Symfony Serializer 6.4.33+, ensuring test cases pass against the new behavior.

**Fix:** https://github.com/apigee/apigee-edge-drupal/issues/1240


shishir-intelli (approved):
LGTM

PR #467, opened 2026-02-05
Ignore 'originalEmail' when creating developer entities

shailesh-google:
Updated the serialization process to ignore 'originalEmail' when creating developers. Adjusted the context for entity creation to ensure 'originalEmail' is not included.


shishir-intelli (approved):
LGTM

Issue #469, opened 2026-03-24
Add support for PHP 8.4 and PHP 8.5


shailesh-google:
To ensure the project remains secure, performant, and up-to-date with the broader PHP ecosystem, we need to add official support for PHP 8.4 and prepare for PHP 8.5. This involves updating our environment requirements, resolving any new deprecations, and ensuring our continuous integration pipeline tests against these newer versions.

PR #478, opened 2026-06-12
Added PHP 8.4 support

shailesh-google:
Fix: #469 

1. setAccessible(true) Method is deprecated Refer Here:-  https://www.php.net/manual/en/reflectionproperty.setaccessible.php

2. Added AMP Library Due to Error with lowest version test in composer analyze command.

3. Psalm Error levels: https://psalm.dev/docs/running_psalm/error_levels/


shishir-intelli (approved):
LGTM

